PR #176, opened 2023-02-07
Fix crates

piobab:


dancreee (approved):

grod220 (approved):

contracts/red-bank/Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [package]
 name          = "mars-red-bank"
+description   = "The Red Bank exposes all interactions that are done between the user and the money market"
```
larry0x, line 3: This sentence doesn't make sense to be

contracts/rewards-collector/osmosis/Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [package]
 name          = "mars-rewards-collector-osmosis"
+description   = "Manages protocol revenue in various coins from the Red Bank"
```
larry0x, line 3:
```suggestion
description   = "A smart contract that manages protocol revenue in various coins from the Red Bank"
```

contracts/red-bank/Cargo.toml
larry0x:
```suggestion
description   = "A smart contract that manages asset deposit, borrowing, and liquidations"
```

larry0x:
@piobab @dancreee It seems that the clicking "commit suggestion" button makes the commit to not be signed, which prevents the PR from being merged.

piobab:
> @piobab @dancreee It seems that the clicking "commit suggestion" button makes the commit to not be signed, which prevents the PR from being merged.

No, it's my fault. I need to debug but my signing disable from time to time. I pushed fix.

PR #179, opened 2023-02-08
Reject unexpected payments; refund the user

larry0x:
- Add `cw_utils::nonpayable` to several Red Bank functions to prevent users from sending coins
- Bump version to v1.1; during 1.0->1.1 migration, refund the 10k axlUSDC to the sender


scripts/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "scripts",
-  "version": "1.0.0",
+  "version": "1.1.0",
+  "license": "GPL-3.0-or-later",
```
larry0x, line 4: Add license here to get rid of `warning package.json: No license field`

Cargo.toml
```diff
@@ -34,7 +34,7 @@ anyhow            = "1.0.68"
 bech32            = "0.9.1"
 cosmwasm-schema   = "1.1.9"
 cosmwasm-std      = "1.1.9"
-cw2               = "1.0.1"
+cw2               = { git = "https://github.com/mars-protocol/cw-plus", rev = "4014255" }
```
grod220, line 37: Why did we fork this repo?

grod220 (commented):
This makes sense to me. However, are we sure refunding is the only action thing that needs to be corrected? Don't some users owe less than they should? Or is that cost borne by the counterparties and those users are getting less rewards?

Cargo.toml
larry0x: Implemented this: https://github.com/CosmWasm/cw-plus/issues/857

larry0x:
@piobab @dancreee I realized that the 10k USDC actually accrues interest while sitting in Red Bank.

This means if we really want to get the number lined up - such that the internally tracked liquidity amount matches what bank module tracks, we need to refund the 10k USDC + the interest accrued during the time.

I wrote the following script:

```bash
redBank="osmo1c3ljch9dfw5kf52nfwpxd2zmj2ese7agnx0p9tenkrryasrle5sqf3ftpg"
usdc="ibc/D189335C6E4A68B513C10AB227BF1C1D38C746766278BA3EEB4FB14124F1D858"

# query the USDC market
market=$(osmosisd q wasm contract-state smart $redBank '{"market":{"denom":"'$usdc'"}}' --output json)
collateralScaled=$(echo $market | jq -r '.data.collateral_total_scaled')
debtScaled=$(echo $market | jq -r '.data.debt_total_scaled')

# query the underlying USDC collateral and debt amounts
collateral=$(osmosisd q wasm contract-state smart $redBank '{"underlying_liquidity_amount":{"denom":"'$usdc'","amount_scaled":"'$collateralScaled'"}}' --output json | jq -r '.data')
debt=$(osmosisd q wasm contract-state smart $redBank '{"underlying_debt_amount":{"denom":"'$usdc'","amount_scaled":"'$debtScaled'"}}' --output json | jq -r '.data')

# this is the amount of available USDC liquidity that red bank *thinks* it has
liquidity=$(( collateral - debt ))

# this is the actual amount of available liquidity tracked by bank module
balance=$(osmosisd q bank balances $redBank --output json | jq -r ".balances[] | select(.denom==\"$usdc\") | .amount")

# the difference, we expect this to be 10,001 USDC + interest
diff=$(( $balance - $liquidity ))

echo $diff
```

Output:

```plain
10001544199
```

We should be refunding 10,001.544199 USDC, compared to the 10,001 USDC the user originally deposited. That is, 0.544199 USDC accrued during the time.

If we only refund 10,001, the rest 0.544199 USDC will remain in the contract, affecting interest rate calculation. Which is of course negligible, but it triggers my perfectionist OCD...

piobab:
@larry0x I understand your "perfectionist OCD" and I think it is good to align contract balance with values saved in state.

contracts/red-bank/tests/test_payment.rs
```diff
@@ -0,0 +1,62 @@
...
+    testing::{mock_env, mock_info},
+    Uint128,
+};
+use cw_utils::PaymentError;
+use helpers::th_setup;
+use mars_red_bank::contract;
+use mars_red_bank_types::red_bank::ExecuteMsg;
+
+/// The Red Bank contract has 6 user-facing functions: deposit, withdraw, borrow,
+/// repay, liquidate, and update_asset_collateral_status; amont these, 3 do not
```
piobab, line 14:
```suggestion
/// repay, liquidate, and update_asset_collateral_status; among these, 3 do not
```

Cargo.toml
```diff
@@ -14,7 +14,7 @@ members = [
 ]
 
 [workspace.package]
-version       = "1.0.0"
+version       = "1.1.0"
```
piobab, line 17: Maybe `1.0.1` ?

larry0x: Agree

piobab:
Closing for now. Moving discussion to https://www.notion.so/delphilabs/Refund-funds-3979d05c4404453eae64f104721cc063

PR #180, opened 2023-02-08
Compute utilization rate using internally-tracked parameters

larry0x:


contracts/red-bank/src/interest_rates.rs
```diff
@@ -278,31 +276,26 @@ pub fn get_updated_liquidity_index(market: &Market, timestamp: u64) -> StdResult
 /// NOTE: For a given block, interest rates should not be updated before updating indexes first
 /// as it should result in wrong indexes
 pub fn update_interest_rates(
-    deps: &DepsMut,
     env: &Env,
     market: &mut Market,
-    liquidity_taken: Uint128,
```
larry0x, line 284: In the previous implementation, where utilization is computed based on the contract's coin balance, we need to subtract the amount of coins that will be sent out.

In this new implementation, we don't need this anymore, because the internally tracked params (`collateral_total_scaled` and `liquidity_index`) have already been properly updated in the `apply_accumulated_interests` function.

contracts/red-bank/src/interest_rates.rs
```diff
@@ -278,31 +276,26 @@ pub fn get_updated_liquidity_index(market: &Market, timestamp: u64) -> StdResult
...
-    denom: &str,
```
larry0x, line 285: We also don't need this `denom`, which is already provided by `market.denom`

contracts/red-bank/src/interest_rates.rs
```diff
@@ -278,31 +276,26 @@ pub fn get_updated_liquidity_index(market: &Market, timestamp: u64) -> StdResult
...
     response: Response,
 ) -> Result<Response, ContractError> {
-    // compute utilization rate
-    let contract_current_balance = deps.querier.query_balance(&env.contract.address, denom)?.amount;
-    if contract_current_balance < liquidity_taken {
-        return Err(ContractError::OperationExceedsAvailableLiquidity {});
-    }
```
larry0x, line 292: We don't really need to check whether "operation exceeds available liquidity". If it does, the bank send message will fail.

contracts/red-bank/tests/test_admin.rs
```diff
@@ -831,22 +833,31 @@ fn update_asset_new_reserve_factor_accrues_interest_rate() {
...
+
+    let asset_initial_collateral = asset_liquidity + asset_initial_debt;
+    let collateral_total_scaled =
+        compute_scaled_amount(asset_initial_collateral, Decimal::one(), ScalingOperation::Ceil)
+            .unwrap();
+
+    let initial_utilization_rate = Decimal::from_ratio(debt_total_scaled, collateral_total_scaled);
+    let borrow_rate = ir_model.get_borrow_rate(initial_utilization_rate).unwrap();
+    let liquidity_rate =
+        ir_model.get_liquidity_rate(borrow_rate, initial_utilization_rate, reserve_factor).unwrap();
```
larry0x, line 847: Previously this test wasn't set up properly: the `collateral_total_scaled` was uninitialized (default to zero), and borrow/liquidity rates were just some made up numbers that do not correspond to the utilization rate. 

Now, in order for the test to pass, we need to actually choose the logical values for these parameters.

I consider this a win - it forces us to improve the quality of our tests!

contracts/red-bank/tests/test_admin.rs
```diff
@@ -886,8 +897,10 @@ fn update_asset_new_reserve_factor_accrues_interest_rate() {
     )
     .unwrap();
     let expected_liquidity = asset_liquidity;
-    let expected_utilization_rate =
-        Decimal::from_ratio(expected_debt, expected_liquidity + expected_debt);
+    // in this particular example, we have to subtract 1 from the total underlying
+    // collateral amount here, because of rounding error.
+    let expected_collateral = expected_liquidity + expected_debt - Uint128::new(1);
+    let expected_utilization_rate = Decimal::from_ratio(expected_debt, expected_collateral);
```
larry0x, line 903: The rounding error happens when converting the collateral's scaled amount to the underlying amount. Here I manually subtract the expected underlying amount by 1 to reflect the error. Not sure if there's a better way.

piobab (approved):

contracts/red-bank/src/interest_rates.rs
grod220: True, but I wonder if this will later make it harder to know why the transaction fails. Error messaging already vague as is in deployed contracts.

grod220 (approved):

PR #181, opened 2023-02-10
MP-2159. stAsset/Asset geometric price source.

piobab:
stATOM oracle we're gonna use the TWAP from the stATOM/ATOM pool
so to standardize to OSMO we're gonna need to do: stATOM/ATOM * ATOM/OSMO
ATOM/OSMO from our current feed

### DIFF
--- a/contracts/oracle/osmosis/src/price_source.rs
+++ b/contracts/oracle/osmosis/src/price_source.rs
@@ -63,6 +63,15 @@ pub struct DowntimeDetector {
     pub recovery: u64,
 }
 
+impl DowntimeDetector {
+    fn fmt(opt_dd: &Option<Self>) -> String {
+        match opt_dd {
+            None => "None".to_string(),
+            Some(dd) => format!("Some({dd})"),
+        }
+    }
+}
+
 impl fmt::Display for DowntimeDetector {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{:?}:{}", self.downtime, self.recovery)
@@ -152,10 +161,7 @@ impl fmt::Display for OsmosisPriceSource {
                 window_size,
                 downtime_detector,
             } => {
-                let dd_fmt = match downtime_detector {
-                    None => "None".to_string(),
-                    Some(dd) => format!("Some({dd})"),
-                };
+                let dd_fmt = DowntimeDetector::fmt(downtime_detector);
                 format!("arithmetic_twap:{pool_id}:{window_size}:{dd_fmt}")
             }
             OsmosisPriceSource::GeometricTwap {
@@ -163,10 +169,7 @@ impl fmt::Display for OsmosisPriceSource {
                 window_size,
                 downtime_detector,
             } => {
-                let dd_fmt = match downtime_detector {
-                    None => "None".to_string(),
-                    Some(dd) => format!("Some({dd})"),
-                };
+                let dd_fmt = DowntimeDetector::fmt(downtime_detector);
                 format!("geometric_twap:{pool_id}:{window_size}:{dd_fmt}")
             }
             OsmosisPriceSource::XykLiquidityToken {
@@ -178,10 +181,7 @@ impl fmt::Display for OsmosisPriceSource {
                 window_size,
                 downtime_detector,
             } => {
-                let dd_fmt = match downtime_detector {
-                    None => "None".to_string(),
-                    Some(dd) => format!("Some({dd})"),
-                };
+                let dd_fmt = DowntimeDetector::fmt(downtime_detector);
                 format!("staked_geometric_twap:{transitive_denom}:{pool_id}:{window_size}:{dd_fmt}")
             }
         };
--- a/contracts/oracle/osmosis/src/price_source.rs
+++ b/contracts/oracle/osmosis/src/price_source.rs
@@ -112,6 +112,30 @@ pub enum OsmosisPriceSource {
     XykLiquidityToken {
         pool_id: u64,
     },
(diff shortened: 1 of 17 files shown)


contracts/oracle/osmosis/src/price_source.rs
```diff
@@ -112,6 +121,30 @@ pub enum OsmosisPriceSource {
...
+    ///
+    /// Example:
+    /// stATOM/OSMO = stATOM/ATOM * ATOM/OSMO
+    /// where:
+    /// - stATOM/ATOM price calculated using the geometric TWAP from the stATOM/ATOM pool.
+    /// - ATOM/OSMO price comes from the Mars Oracle contract.
+    ///
+    /// NOTE: `pool_id` must point to stAsset/Asset Osmosis pool.
+    /// Asset/OSMO price source should be available in the Mars Oracle contract.
+    StakedGeometricTwap {
```
piobab, line 134: I am open for different naming.

contracts/oracle/osmosis/src/price_source.rs
```diff
@@ -112,6 +121,30 @@ pub enum OsmosisPriceSource {
...
+        /// Transitive denom for which we query price in OSMO
+        transitive_denom: String,
```
piobab, line 136: Used explicit denom here in order to avoid quering pool and unwrapping denom from the assets vector during price querying.

contracts/oracle/osmosis/src/price_source.rs
```diff
@@ -63,6 +63,15 @@ pub struct DowntimeDetector {
...
 }
 
+impl DowntimeDetector {
+    fn fmt(opt_dd: &Option<Self>) -> String {
+        match opt_dd {
+            None => "None".to_string(),
+            Some(dd) => format!("Some({dd})"),
+        }
+    }
+}
```
larry0x, line 73: What's the purpose of this when we already have the `fmt::Display` trait?

contracts/oracle/osmosis/src/price_source.rs
```diff
@@ -128,26 +161,29 @@ impl fmt::Display for OsmosisPriceSource {
                 window_size,
                 downtime_detector,
             } => {
-                let dd_fmt = match downtime_detector {
-                    None => "None".to_string(),
-                    Some(dd) => format!("Some({dd})"),
-                };
+                let dd_fmt = DowntimeDetector::fmt(downtime_detector);
                 format!("arithmetic_twap:{pool_id}:{window_size}:{dd_fmt}")
```
larry0x, line 165: Off topic: I'm thinking that instead of manually implementing `fmt::Display`, we can perhaps do

```rust
let price_source_str = serde_json_wasm::to_string(&price_source)?;
```

contracts/oracle/osmosis/src/price_source.rs
```diff
@@ -194,6 +230,16 @@ impl PriceSource<Empty> for OsmosisPriceSource {
...
             }
+            OsmosisPriceSource::StakedGeometricTwap {
+                transitive_denom,
+                pool_id,
+                window_size,
+                downtime_detector,
+            } => {
+                let pool = query_pool(querier, *pool_id)?;
+                helpers::assert_osmosis_pool_assets(&pool, denom, transitive_denom)?;
+                helpers::assert_osmosis_twap(*window_size, downtime_detector)
```
larry0x, line 241: Should we also assert that the contract has the price source for `transitive_denom`?

larry0x (approved):
LGTM

contracts/oracle/osmosis/src/price_source.rs
piobab: private function to replace duplicated code in 3 places

contracts/oracle/osmosis/src/price_source.rs
piobab: hmm we can but in that case would be good to the same for LPtoken (to be consistent).

larry0x: I thought LP token does this... Anyways, adding a new price source is controlled by governance, so not a bit deal if we don't do the verification.

piobab: Ideally, we have protection for removal too if a price source depends on another one. I am not sure if it worth to implement right now (as you said `is controlled by governance`).

contracts/oracle/osmosis/src/price_source.rs
grod220: Could you do `impl Display for &Option<DowntimeDetector>` and just use `to_string()` places instead?

contracts/oracle/osmosis/src/price_source.rs
grod220: Is transitive_denom the same as the unwrapped/unstaked denom?

grod220: I think we need a bit more info in the comment about what a transitive denom is

grod220 (approved):

contracts/oracle/osmosis/src/price_source.rs
piobab: stAsset/Asset * Asset/Osmo
Asset here =  transitive_denom

contracts/oracle/osmosis/src/price_source.rs
piobab: Unfortunatelly I can't :(

contracts/oracle/osmosis/src/price_source.rs
piobab: Added

dancreee (approved):

PR #183, opened 2023-02-24
Pyth

piobab:


Cargo.toml
```diff
@@ -42,6 +42,7 @@ mars-owner        = "1.0.0"
 osmosis-std       = "0.14.0"
 osmosis-test-tube = "14.1.1"
 prost             = { version = "0.11.5", default-features = false, features = ["prost-derive"] }
+pyth-sdk-cw       = "1.0.0"
```
grod220, line 45: lol they missed the opportunity to join the `cw-` party if they did `cw-pyth-sdk`

contracts/oracle/base/src/contract.rs
```diff
@@ -157,13 +171,42 @@ where
...
+        let mut config = self.config.load(deps.storage)?;
+        config.base_denom = base_denom.unwrap_or(config.base_denom);
+        self.config.save(deps.storage, &config)?;
+
+        let mut pyth_cfg = self.pyth_config.load(deps.storage)?;
+        pyth_cfg.pyth_contract_addr =
+            option_string_to_addr(deps.api, pyth_contract_addr, pyth_cfg.pyth_contract_addr)?;
+        self.pyth_config.save(deps.storage, &pyth_cfg)?;
+
+        let response = Response::new().add_attribute("action", "update_config");
```
grod220, line 196: Should we add keys for the updated fields?

contracts/oracle/base/src/contract.rs
```diff
@@ -202,6 +245,7 @@ where
 
     fn query_price(&self, deps: Deps<C>, env: Env, denom: String) -> ContractResult<PriceResponse> {
```
grod220, line 246: Almost think this function should simply call `query_prices(limit: 1).first()`

contracts/oracle/osmosis/src/helpers.rs
```diff
@@ -71,3 +72,22 @@ pub fn assert_osmosis_twap(
 
     Ok(())
 }
+
+/// Assert Pyth configuration
+pub fn assert_pyth(max_confidence: Decimal, max_deviation: Decimal) -> ContractResult<()> {
+    // TODO: update validation once confirmed with Risk team
```
grod220, line 78: Should confidence & max_deviation be config variables instead?

contracts/oracle/osmosis/src/price_source.rs
```diff
@@ -390,13 +441,13 @@ impl OsmosisPriceSource {
     fn query_staked_asset_price(
         deps: &Deps,
         env: &Env,
-        denom: &str,
-        transitive_denom: &str,
-        base_denom: &str,
+        denoms: (&str, &str, &str),
```
grod220, line 444: I'm not convinced this is better. It's now less documented for consumers.

contracts/oracle/osmosis/src/price_source.rs
```diff
@@ -413,10 +464,95 @@ impl OsmosisPriceSource {
...
+        env: &Env,
+        price_feed_id: PriceIdentifier,
+        max_staleness: u64,
+        max_confidence: Decimal,
+        max_deviation: Decimal,
+        pyth_config: &PythConfig,
+    ) -> ContractResult<Decimal> {
+        let current_time = env.block.time.seconds();
+
+        let price_feed_response: PriceFeedResponse =
```
grod220, line 484: Do you have to manually type `PriceFeedResponse`? Their library doesn't do that?

contracts/oracle/osmosis/src/price_source.rs
```diff
@@ -413,10 +464,95 @@ impl OsmosisPriceSource {
...
+        let price_feed = price_feed_response.price_feed;
+
+        // Get the current price and confidence interval from the price feed
+        let current_price = price_feed.get_price_unchecked();
+
+        // Check if the current price is not too old
+        if (current_time - current_price.publish_time as u64) > max_staleness {
+            return Err(InvalidPrice {
+                reason: format!(
+                    "current price timestamp is too old/stale. published: {}, now: {}",
```
grod220, line 495: Instead of `timestamp` I think we should put `publish time`. Same below.

contracts/oracle/osmosis/src/price_source.rs
```diff
@@ -149,6 +151,24 @@ pub enum OsmosisPriceSource {
...
+        /// The maximum number of seconds since the last price was by an oracle, before
+        /// rejecting the price as too stale
+        max_staleness: u64,
+
+        /// The maximum confidence deviation allowed for an oracle price.
+        ///
+        /// The confidence is measured as the percent of the confidence interval
+        /// value provided by the oracle as compared to the weighted average value
+        /// of the price.
+        max_confidence: Decimal,
```
grod220, line 167: I'd consider renaming this to `max_confidence_dev`.

contracts/oracle/osmosis/src/price_source.rs
```diff
@@ -413,10 +464,95 @@ impl OsmosisPriceSource {
...
+        }
+
+        let current_price_dec = scale_to_exponent(current_price.price as u128, current_price.expo)?;
+        let ema_price_dec = scale_to_exponent(ema_price.price as u128, ema_price.expo)?;
+
+        // Check confidence deviation
+        let confidence = scale_to_exponent(current_price.conf as u128, current_price.expo)?;
+        if confidence.checked_div(ema_price_dec)? > max_confidence {
+            return Err(InvalidPrice {
+                reason: "price confidence exceeding max".to_string(),
```
grod220, line 528: I'd do `"price confidence deviation exceeds max allowed"`

contracts/oracle/osmosis/src/price_source.rs
```diff
@@ -413,10 +464,95 @@ impl OsmosisPriceSource {
...
+            });
+        }
+
+        // Check price deviation
+        let delta = if current_price_dec > ema_price_dec {
+            current_price_dec - ema_price_dec
+        } else {
+            ema_price_dec - current_price_dec
+        };
+        if delta.checked_div(ema_price_dec)? > max_deviation {
```
grod220, line 538: You could also do:
```rust
let delta = abs(current_price_dec - ema_price_dec)
```

contracts/oracle/osmosis/src/price_source.rs
```diff
@@ -413,10 +464,95 @@ impl OsmosisPriceSource {
...
+            return Err(InvalidPrice {
+                reason: "price deviation exceeding max".to_string(),
+            });
```
grod220, line 541: Would be nice to put the deviation in the error message.

contracts/oracle/osmosis/src/price_source.rs
```diff
@@ -413,10 +464,95 @@ impl OsmosisPriceSource {
...
+        }
+
+        Ok(current_price_dec)
+    }
+}
+
+fn scale_to_exponent(value: u128, expo: i32) -> ContractResult<Decimal> {
```
grod220, line 548: Instead of the `as u128` maybe you can change this signature to `value: Impl<Into<u128>>`. Or does that work? 🤔

grod220: Also, what's the purpose of this? Maybe some code comments would help. Is it to convert the confidence to a Decimal?

grod220 (commented):
Did you consider using multi-test to test out the Pyth contract?

contracts/oracle/base/src/contract.rs
piobab: Added

contracts/oracle/base/src/contract.rs
piobab: I don't see advantage of doing this.

contracts/oracle/osmosis/src/helpers.rs
piobab: IMO it should be per asset because different asset could have different risk etc. More elastic for Risk team.

contracts/oracle/osmosis/src/price_source.rs
piobab: Yeah, but clippy warns about too much params. I can leave prev version but then we need
`#[allow(clippy::too_many_arguments)]`

contracts/oracle/osmosis/src/price_source.rs
piobab: It does, copy paste from Pyth sample contract https://github.com/pyth-network/pyth-crosschain/blob/main/target_chains/cosmwasm/examples/cw-contract/src/contract.rs#L93

contracts/oracle/osmosis/src/price_source.rs
piobab: Updated

contracts/oracle/osmosis/src/price_source.rs
piobab: hmmm I leave it for now as it is.

contracts/oracle/osmosis/src/price_source.rs
piobab: Instead of the as u128 maybe you can change this signature to value: Impl<Into<u128>>. Or does that work?
 - no, it doesn't make sense because Pyth reports prices in i64 so firstly it is better to cast to u128.

Also, what's the purpose of this? Maybe some code comments would help. Is it to convert the confidence to a Decimal?
 - I added comment. In general Pyth responses are in primitive types so I think this is the way we should convert them to Cosmwasm types.
Anyway, if you have some suggestion LMK.

contracts/oracle/osmosis/src/price_source.rs
grod220: I'd prefer having named arguments vs the tuple, but will leave this up to you

contracts/oracle/osmosis/src/price_source.rs
grod220: The comment definitely helps. Pyth should consider just using the Decimal struct themselves.

grod220 (approved):

contracts/oracle/osmosis/src/price_source.rs
larry0x: Agree with Gabe, prefer named arguments.

We can create a `clippy.toml` where we customized the max argument number or turn it off completely.

contracts/oracle/base/src/contract.rs
```diff
@@ -202,6 +252,7 @@ where
...
         let cfg = self.config.load(deps.storage)?;
+        let pyth_cfg = self.pyth_config.load(deps.storage)?;
```
larry0x, line 255: By doing this, even assets that don't use Pyth (e.g. OSMO) need to load the pyth config, which can incur pretty big gas costs. If a user has 10 assets in their position, this config needs to be loaded 10 times. Can we optimize this? Perhaps put pyth config inside the price source?

contracts/oracle/osmosis/src/price_source.rs
piobab: Updated with 'allow'

contracts/oracle/base/src/contract.rs
piobab: Ok, so we could do:
1) include pyth_contract_addr in every Pyth price source
2) move pyth_contract_addr from separate state to config (which is already loaded too)

cc @grod220

piobab: wdyt @larry0x ?

larry0x (approved):

PR #191, opened 2023-04-25
Mp 2487 extract asset params

piobab:
Use `mars-params` contract https://github.com/mars-protocol/mars-common/tree/main/contracts/params

base branch should be changed. We should create new release branch but for now it is ok.


piobab:
Naming proposition for params in Red Bank and `params` contract are more than welcome. Both of them create full asset configuration.

contracts/red-bank/src/execute.rs
```diff
@@ -42,23 +42,16 @@ pub fn instantiate(deps: DepsMut, msg: InstantiateMsg) -> Result<Response, Contr
...
+    if address_provider.is_none() {
         return Err(MarsError::InstantiateParamsUnavailable {}.into());
     };
 
     let config = Config {
         address_provider: option_string_to_addr(deps.api, address_provider, zero_address())?,
-        close_factor: close_factor.unwrap(),
     };
 
-    config.validate()?;
```
grod220, line 60: No more validation?

contracts/red-bank/src/execute.rs
```diff
@@ -643,7 +563,7 @@ pub fn borrow(
 
         let debt_amount_scaled = borrower.debt_amount_scaled(deps.storage, &denom)?;
 
-        let asset_market = MARKETS.load(deps.storage, &denom)?;
+        let asset_market = MARKETS.load(deps.storage, &denom)?; // TODO do we need this one?
```
grod220, line 566: Shouldn't there no longer be "red bank markets"? As everything lives in the assets params contract.

packages/types/src/red_bank/market.rs
```diff
@@ -40,13 +29,6 @@ pub struct Market {
     pub collateral_total_scaled: Uint128,
     /// Total debt scaled for the market's currency
     pub debt_total_scaled: Uint128,
```
grod220, line 31: What are the thoughts for keeping this in Red Bank? Is it because they need to be constantly updated? Also, is `Market` still the right word for this struct?

contracts/red-bank/src/execute.rs
piobab: Address is already validated.

contracts/red-bank/src/execute.rs
piobab: Hmmm thinking about some naming.. we keep indices, amount of collateral / debt etc. Do you think about changing this?

packages/types/src/red_bank/market.rs
piobab: IMO better to leave it in RB. Otherwise we will have part data in external contract (related to global amount of asset collateral and debt) and the rest (for every user) in RB.
Even if we decided to move it I would say it is overoptimization in current state of the project.

contracts/red-bank/src/query.rs
```diff
@@ -15,25 +15,19 @@ use crate::{
...
 const DEFAULT_LIMIT: u32 = 5;
 const MAX_LIMIT: u32 = 10;
 
 pub fn query_config(deps: Deps) -> StdResult<ConfigResponse> {
     let owner_state = OWNER.query(deps.storage)?;
-    let emergency_owner_state = EMERGENCY_OWNER.query(deps.storage)?;
     let config = CONFIG.load(deps.storage)?;
     Ok(ConfigResponse {
         owner: owner_state.owner,
         proposed_new_owner: owner_state.proposed,
```
grod220, line 29: Think it's best to return the entire owner state object instead of just those two fields

grod220 (approved):

brimigs (approved):
LGTM :)

PR #192, opened 2023-05-06
new testnet deployment

brimigs:
update deploy scripts + config file for osmo-test-5


scripts/deploy/base/index.ts
```diff
@@ -1,7 +1,19 @@
...
+  osmoAsset,
+  atomOracle,
+  axlUSDCAsset,
+  axlUSDCOracle,
+  atomAssetTest,
+  axlUSDCAssetTest,
+  axlUSDCOracleTest,
+  osmoOracle,
+  marsAssetTest,
+  marsOracleTest,
```
dancreee, line 15: probably a cleaner way to do this then importing both mainnet and testnet assets into same file but ok for now

dancreee (approved):

scripts/deploy/osmosis/config.ts
```diff
@@ -208,16 +234,78 @@ export const axlUSDCAsset: AssetConfig = {
...
+
 export const atomOracle: OracleConfig = {
-  denom: 'ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2',
+  denom: atom,
   pool_id: 1,
   window_size: 1800,
   downtime_detector: { downtime: 'duration30m', recovery: 7200 },
 }
 
+// export const atomOracleTest: OracleConfig = {
```
piobab, line 285: To be removed?

piobab (approved):

PR #193, opened 2023-05-23
Pyth v1

piobab:


brimigs (approved):

Cargo.toml
```diff
@@ -14,7 +14,7 @@ members = [
 ]
 
 [workspace.package]
-version       = "1.0.1"
+version       = "1.0.2"
```
grod220, line 17: Doesn't Pyth upgrade merit a minor version release? `1.1`?

grod220 (approved):

Cargo.toml
piobab: Yeah, good point. Updated.

PR #194, opened 2023-05-24
Update decimals

piobab:
We have to introduce denomination to something smaller, let’s say “uusd” - representing USD with 6 decimals.

1. Calculate uosmo/uusd

`osmo_price_in_usd = 0.59958994`

`calculate how much 1 uosmo is worth in uusd`

`1 OSMO = 0.59958994 USD`

`1 OSMO = 1000000 uosmo`

`1 USD = 1000000 uusd`

`1000000 uosmo = 0.59958994 * 1000000 uusd`

`1 uosmo = 0.59958994 uusd`

1. Calculate ueth/uusd

`eth_price_in_usd = 1815.98000001`

`1 ETH = 1815.98000001 USD`

`1 ETH = 10^18 ueth`

`1 USD = 10^6 uusd`

`10^18 ueth = 1815.98000001 * 10^6 uusd`

`1 ueth = 1815.98000001 * 10^6 uusd * 10^(-18) = 1815.98000001 * 10^(-12) usd = 0.00000000181598 usd`

Collaterals:

120 OSMO

88.55 ETH

Total amount of collateral in USD: 

120 * `0.59958994` + 88.55 * `1815.98000001` = 160876.979794 USD

Calculate SC way using utoken amounts and uusd price (1 USD = 1000000 uusd):

120000000 uosmo * `0.59958994` + 88550000000000000000 ueth * `0.00000000181598` = 160876979793 uusd

160876979793 uusd / 10^6 = 160876.979793 USD

We have to decide how much decimals we want to use:

If we use 6 then:

```markup
let ueth_price_in_usd = scale_pyth_price(181598000001u128, -8, 18u8).unwrap();
assert_eq!(ueth_price_in_usd, Decimal::from_str("0.00000000181598").unwrap());
```

If we use 8 then:

```markup
let ueth_price_in_usd = scale_pyth_price(181598000001u128, -8, 18u8).unwrap();
assert_eq!(ueth_price_in_usd, Decimal::from_str("0.000000181598000001").unwrap());
```
IMO 6 decimals should be ok.

**We introduce:**

- **decimals for `base_denom`**
- **decimals for requested denom price**

**Frontend / backend needs to be aware that the prices are denominated in uusd and convert back to USD if needed**


grod220 (approved):

integration-tests/tests/test_oracles.rs
```diff
@@ -57,6 +57,7 @@ fn querying_xyk_lp_price_if_no_price_for_tokens() {
         &InstantiateMsg {
             owner: signer.address(),
             base_denom: "uosmo".to_string(),
+            base_denom_decimals: 6u8,
         },
```
apollo-sturdy, line 61: Just a thought. Here we have the same instantiate message once in each test in this file. Perhaps to avoid big diffs like this in the future we can create a function to generate it, or perhaps even impl Default trait on InstantiateMsg?

contracts/oracle/base/src/traits.rs
```diff
@@ -28,9 +29,10 @@ where
     ///
     /// - `denom`: The coin whose price is to be queried.
     ///
-    /// - `base_denom`: The coin in which the price is to be denominated in.
+    /// - `config.base_denom`: The coin in which the price is to be denominated in.
     ///   For example, if `denom` is uatom and `base_denom` is uosmo, the
     ///   function should return how many uosmo is per one uatom.
+    /// - `config.base_denom_decimals`: The coin decimals used for price normalization.
```
apollo-sturdy, line 35: This comment could be more clear. If I'm new to the project I probably don't understand it.

brimigs (approved):

contracts/oracle/osmosis/src/price_source.rs
```diff
@@ -362,10 +372,12 @@ impl PriceSourceUnchecked<OsmosisPriceSourceChecked, Empty> for OsmosisPriceSour
                 contract_addr,
                 price_feed_id,
                 max_staleness,
+                denom_decimals,
             } => Ok(OsmosisPriceSourceChecked::Pyth {
                 contract_addr: deps.api.addr_validate(contract_addr)?,
                 price_feed_id: *price_feed_id,
                 max_staleness: *max_staleness,
+                denom_decimals: *denom_decimals,
             }),
```
apollo-sturdy, line 381: Are we able to validate that the assets of the price feed are the correct ones? If the denomination asset of the value pyth sends back is not equal to the `base_denom` then everything will be messed up. Would be good to validate if we can.

apollo-sturdy (commented):
Math looks correct to me. But now reading this I'm not so sure that adding decimals for base_denom is the best idea as it is now only used for conversion with USD, which assumes base_denom is scaled USD? In fact, with the addition of this implementation, base_denom must be some scaled version of whatever denom all added pyth prices use. Maybe a better idea would be to do like we did for Astroport and have `route_assets`? This way we can support any base denom and also any pyth price feed that is not denominated in USD.

This way you could still have `base_denom` be for example `uusd` and you would have for pyth assets `route_assets = ['USD']`, were price source for `USD/uusd` would be `0.000001`.

contracts/oracle/base/src/contract.rs
```diff
@@ -180,13 +183,17 @@ where
 
         let mut config = self.config.load(deps.storage)?;
         let prev_base_denom = config.base_denom.clone();
+        let prev_base_denom_decimals = config.base_denom_decimals;
```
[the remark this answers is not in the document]
piobab: Because it is overriden in next line:
`config.base_denom_decimals = base_demom_decimals.unwrap_or(config.base_denom_decimals);`

piobab:
> `assumes base_denom is scaled USD`
Correct.

Not sure how we could incorporate route_assets.. This is not the case that we convert something to different denom (out denom) but mostly do scaling to incorporate decimals correctly.

Cargo.toml
```diff
@@ -34,7 +34,7 @@ anyhow            = "1.0.68"
 bech32            = "0.9.1"
 cosmwasm-schema   = "1.1.9"
 cosmwasm-std      = "1.1.9"
-cw2               = { git = "https://github.com/mars-protocol/cw-plus", rev = "4014255" }
+cw2               = { git = "https://github.com/CosmWasm/cw-plus", rev = "de1fb0b" }
```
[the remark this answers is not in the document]
piobab: IMO it is not big problem, git dependency is proper way but for sure crates is better. For now at least we have commit id.

contracts/oracle/osmosis/src/price_source.rs
apollo-sturdy: Too bad. Could maybe add to doc comments of Pyth variant of price source that we need to be very careful when adding it

contracts/oracle/base/src/traits.rs
piobab: Will be removed once I introduce routes assets

piobab:
Updated PR:
- use wasms from rust-optimizer for integration tests (new Rust breaks uploading wasm from `target` directory)
```
ExecuteError { msg: "failed to execute message; message index: 0: Error calling the VM: Error during static Wasm validation: Wasm bytecode could not be deserialized. Deserialization error: \"Unknown opcode 192\": create wasm contract failed" }
```
- added `sudo chown -R $USER .` in coverage workflow

PR #199, opened 2023-06-05
feat: Add basic support for multiple incentive denoms

apollo-sturdy:


contracts/incentives/src/error.rs
```diff
@@ -27,4 +27,13 @@ pub enum ContractError {
     InvalidIncentive {
         reason: String,
     },
+
+    #[error("Invalid Pagination Params. If start_after_incentive_denom is supplied, then start_after_collateral_denom must also be supplied")]
+    InvalidPaginationParams,
```
larry0x, line 32: Why not define the pagination params as a struct, this way you no longer need to worry about this error:

```rust
struct PaginationParams {
    pub incentive_denom: String,
    pub collateral_denom: String,
}

enum QueryMsg {
    AssetIncentives {
        start_after: Option<PaginationParams>,
        limit: Option<u32>,
    },
}
```

contracts/incentives/src/state.rs
```diff
@@ -1,13 +1,121 @@
...
 pub const CONFIG: Item<Config> = Item::new("config");
 
-// namespaces (for buckets)
-pub const ASSET_INCENTIVES: Map<&str, AssetIncentive> = Map::new("incentives");
-pub const USER_ASSET_INDICES: Map<(&Addr, &str), Decimal> = Map::new("indices");
-pub const USER_UNCLAIMED_REWARDS: Map<&Addr, Uint128> = Map::new("unclaimed_rewards");
+/// A map containing a configuration of an incentive for a given collateral and incentive denom.
+/// The key is (collateral denom, incentive denom).
+/// We use String instead of &str for the key because Map::prefix_range gives borrow issues with &str.
+pub const ASSET_INCENTIVES: Map<(String, String), AssetIncentive> = Map::new("incentives");
```
larry0x, line 17: This can be `Map<(&str, &str), AssetIncentive>` (using refs instead of owned values as the key) which can avoid several clones

contracts/incentives/src/contract.rs
```diff
@@ -296,116 +315,153 @@ pub fn execute_balance_change(
...
+                USER_UNCLAIMED_REWARDS.update(
+                    deps.storage,
+                    (&user_addr, &collateral_denom, &incentive_denom),
+                    |ur: Option<Uint128>| -> StdResult<Uint128> {
+                        match ur {
+                            Some(unclaimed_rewards) => Ok(unclaimed_rewards + accrued_rewards),
+                            None => Ok(accrued_rewards),
+                        }
+                    },
+                )?;
```
larry0x, line 368: This can perhaps be extracted to a helper function to improve readability?

contracts/incentives/src/contract.rs
```diff
@@ -296,116 +315,153 @@ pub fn execute_balance_change(
...
+
+            user_asset_index_key.save(deps.storage, &asset_incentive.index)?;
         }
 
-        user_asset_index_key.save(deps.storage, &asset_incentive.index)?;
+        event = event
+            .add_attribute(format!("rewards_accrued_{}", incentive_denom), accrued_rewards)
+            .add_attribute(
+                format!("asset_index_{}", incentive_denom),
+                asset_incentive.index.to_string(),
```
larry0x, line 378: I'm not sure if it's a good practice to put information in the keys. I'd suggest consult backend guys on what they think

contracts/incentives/src/error.rs
larry0x: Ok I see, what it means is
- it's ok to only supply `collateral_denom` but not supply `incentive_denom`
- it's ok to supply both or supply neither
- it's NOT ok supply only `incentive_denom` but not `collateral_denom`

In this case you can probably do

```rust
struct PaginationParams {
    pub collateral_denom: String,
    pub incentive_denom: Option<String>,
}

enum QueryMsg {
    AssetIncentives {
        start_after: Option<PaginationParams>,
        limit: Option<u32>,
    },
}
```

Not sure if this is better than the current approach tho

contracts/incentives/src/state.rs
apollo-sturdy: See the code comment which explains why I used `String` instead of `&str`. Maybe you have a way around it? If you try changing it to &str you will get compilation error in `fn paginate_asset_incentives`, but maybe there is some other workaround.

contracts/incentives/src/error.rs
apollo-sturdy: Yeah not sure if thats better. Can change it if someone has a strong preference.

contracts/incentives/src/contract.rs
apollo-sturdy: Yeah I wasn't really sure here since we don't know at compile time which denoms are being used. The other option would be to emit a separate event for each denom. Maybe that's better?

contracts/incentives/src/error.rs
apollo-sturdy: Also curious if you agree with this pagination method. I suppose it maybe has to be this way to support pagination down to single item granularity since the key is `(collateral_denom, incentive_denom)`?

contracts/incentives/src/contract.rs
elfedy: @apollo-sturdy as a rule of a thumb for events, it is desired to always have the same keys given the same event, else an indexer has to start guessing the structure of the event before parsing the keys which is not ideal.

Here you have two options, either parameterize the attributes as arrays so `rewards_accrued: 1000denoma, 222000denomb` or whatever depending on the value. Or just emit a new event per denom. I think the latter is better in this case, at least from an indexing perspective.

contracts/incentives/src/contract.rs
apollo-sturdy: Fixed

contracts/incentives/src/contract.rs
apollo-sturdy: Fixed.

contracts/incentives/src/state.rs
piobab: @apollo-sturdy please take a look https://github.com/mars-protocol/red-bank/pull/new/avoid-clone-keys

contracts/incentives/src/contract.rs
```diff
@@ -296,116 +315,156 @@ pub fn execute_balance_change(
...
+                deps.as_ref(),
+                &env.block,
+                &red_bank_addr,
+                &user_addr,
+                &collateral_denom,
+                &incentive_denom,
+            )?;
+
+        // Commit updated asset_incentives and user indexes
+        if let Some(user_asset_incentive_status) = user_asset_incentive_statuses_to_update {
```
piobab, line 421:
```
        if let Some(UserAssetIncentiveStatus {
            user_index_current,
            asset_incentive_updated,
        }) = user_asset_incentive_statuses_to_update
        {
            ASSET_INCENTIVES.save(
                deps.storage,
                (collateral_denom.clone(), incentive_denom.clone()),
                &asset_incentive_updated,
            )?;

            if asset_incentive_updated.index != user_index_current {
                USER_ASSET_INDICES.save(
                    deps.storage,
                    (&user_addr, &collateral_denom, &incentive_denom),
                    &asset_incentive_updated.index,
                )?
            }
        }
```

contracts/incentives/src/contract.rs
```diff
@@ -296,116 +315,156 @@ pub fn execute_balance_change(
...
+            deps.storage,
+            (&user_addr, &collateral_denom, &incentive_denom),
+            &Uint128::zero(),
+        )?;
 
-    let mut response = Response::new();
-    if !total_unclaimed_rewards.is_zero() {
-        let config = CONFIG.load(deps.storage)?;
-        // Build message to send mars to the user
+        if !unclaimed_rewards.is_zero() {
```
piobab, line 446:
```
total_unclaimed_rewards
                .entry(incentive_denom)
                .and_modify(|counter| *counter += unclaimed_rewards)
                .or_insert(unclaimed_rewards);
```

contracts/incentives/src/contract.rs
```diff
@@ -472,36 +548,80 @@ pub fn query_config(deps: Deps) -> StdResult<ConfigResponse> {
...
+    for ((collateral_denom, incentive_denom), _) in asset_incentives {
+        let (unclaimed_rewards, _) = compute_user_unclaimed_rewards(
+            deps,
+            &env.block,
+            &red_bank_addr,
+            &user_addr,
+            &collateral_denom,
+            &incentive_denom,
+        )?;
+        if let Some(x) = total_unclaimed_rewards.get_mut(&incentive_denom) {
```
piobab, line 611:
```
total_unclaimed_rewards
                .entry(incentive_denom)
                .and_modify(|counter| *counter += unclaimed_rewards)
                .or_insert(unclaimed_rewards);
```

contracts/incentives/src/error.rs
piobab: We could simplify API by requirement to query "incentives" for specific `collateral` and then caller can execute separate msgs.

contracts/incentives/src/helpers.rs
```diff
@@ -92,64 +89,62 @@ pub fn compute_user_unclaimed_rewards(
...
+
+    // Get asset user balances and total supply
+    let collateral: red_bank::UserCollateralResponse = deps.querier.query_wasm_smart(
+        red_bank_addr,
+        &red_bank::QueryMsg::UserCollateral {
+            user: user_addr.to_string(),
+            denom: collateral_denom.to_string(),
+        },
+    )?;
+    let market: red_bank::Market = deps.querier.query_wasm_smart(
```
piobab, line 110: Hmm analyzing this now and if incentives is set for market which is not set in redbank it could fail the whole tx.

contracts/incentives/src/state.rs
apollo-sturdy: > @apollo-sturdy please take a look https://github.com/mars-protocol/red-bank/pull/new/avoid-clone-keys

Aahh perfect. `as_ref()` is what I was missing.

contracts/incentives/src/helpers.rs
apollo-sturdy: Good point. Should we verify that market exists in Red Bank in ExecuteMsg::SetIncentive? It's not possible to remove a market in red bank is it?

contracts/incentives/src/state.rs
```diff
@@ -1,13 +1,143 @@
...
+            }
+        },
+    )?;
+    Ok(())
+}
+
+/// Returns an iterator over all asset incentives, with optional pagination.
+/// Caller should make sure that if start_after_incentive_denom is supplied, then
+/// start_after_collateral_denom is also supplied.
+pub fn paginate_asset_incentives(
```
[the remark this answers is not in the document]
apollo-sturdy: I guess really it's up to frontend devs what they prefer. The downside of your suggestion is that they would need to know all collaterals that the user has incentives for. I could imagine that this is hard to keep track of. They can't just query Red Bank since the user can remove collateral and still have unclaimed rewards.

contracts/incentives/src/helpers.rs
piobab: No, it is not possible to remove. IMO we should verify in SetIncentive that collateral exists in Red Bank.

piobab (approved):

contracts/incentives/src/helpers.rs
apollo-sturdy: Fixed

contracts/incentives/src/state.rs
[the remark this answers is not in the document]
apollo-sturdy: @piobab what do you think? Personally I think current method is better so backend doesn't need to keep track of state

PR #202, opened 2023-06-07
Improve pyth base dec

piobab:
Some inspiration from @apollo-sturdy review in previous PR.
```
Math looks correct to me. But now reading this I'm not so sure that adding decimals for base_denom is the best idea as it is now only used for conversion with USD, which assumes base_denom is scaled USD? In fact, with the addition of this implementation, base_denom must be some scaled version of whatever denom all added pyth prices use. Maybe a better idea would be to do like we did for Astroport and have route_assets? This way we can support any base denom and also any pyth price feed that is not denominated in USD.

This way you could still have base_denom be for example uusd and you would have for pyth assets route_assets = ['USD'], were price source for USD/uusd would be 1000000.
```

This way we can convert pyth USD to different `base_denom`. It can be uusd, uosmo, untrn etc. We only need price source for USD.
IMO `route_assets` don't make sense because all our prices are denominated in the same denom so having route like ["USD", "UOSMO"] (assume that base_denom=uosmo) could end up multiplying `pyth_price_in_USD * usd_price_in_uosmo * uosmo_in_uosmo`.

### DIFF
--- a/contracts/oracle/base/src/contract.rs
+++ b/contracts/oracle/base/src/contract.rs
@@ -73,7 +73,6 @@ where
             deps.storage,
             &Config {
                 base_denom: msg.base_denom,
-                base_denom_decimals: msg.base_denom_decimals,
             },
         )?;
 
@@ -97,8 +96,7 @@ where
             } => self.remove_price_source(deps, info.sender, denom),
             ExecuteMsg::UpdateConfig {
                 base_denom,
-                base_denom_decimals,
-            } => self.update_config(deps, info.sender, base_denom, base_denom_decimals),
+            } => self.update_config(deps, info.sender, base_denom),
         }
     }
 
@@ -173,7 +171,6 @@ where
         deps: DepsMut<C>,
         sender_addr: Addr,
         base_denom: Option<String>,
-        base_demom_decimals: Option<u8>,
     ) -> ContractResult<Response> {
         self.owner.assert_owner(deps.storage, &sender_addr)?;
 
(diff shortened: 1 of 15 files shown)


grod220 (approved):
Nice job, looks simpler now 👍

apollo-sturdy (approved):
> IMO route_assets don't make sense because all our prices are denominated in the same denom so having route like ["USD", "UOSMO"] (assume that base_denom=uosmo) could end up multiplying pyth_price_in_USD * usd_price_in_uosmo * uosmo_in_uosmo.

I don't see how this is a problem, since `uosmo_in_uosmo = 1`. Route assets vec would only be needed if we ever wanted to support a Pyth price not denominated in USD. Right now it looks like only forex pairs are non-USD denominated in Pyth. So maybe its not worth the complexity if we think we will never want it. But on the other hand the code is already there for Astroport route assets so should be able to use it with minimal modifications.

piobab:
> > IMO route_assets don't make sense because all our prices are denominated in the same denom so having route like ["USD", "UOSMO"] (assume that base_denom=uosmo) could end up multiplying pyth_price_in_USD * usd_price_in_uosmo * uosmo_in_uosmo.
> 
> I don't see how this is a problem, since `uosmo_in_uosmo = 1`. Route assets vec would only be needed if we ever wanted to support a Pyth price not denominated in USD. Right now it looks like only forex pairs are non-USD denominated in Pyth. So maybe its not worth the complexity if we think we will never want it. But on the other hand the code is already there for Astroport route assets so should be able to use it with minimal modifications.

IMO it doesn't make sense to have vector of assets. It is ok to have one asset ala "transitive denom". Vector of assets make sense if we want to do something like AXL/OSMO * OSMO/USD * USD/NTRN. Anyway your point is valid once we want to support different denomination than USD but I think it is not in sight.

PR #222, opened 2023-06-26
feat: Add minimum epoch duration check

apollo-sturdy:


contracts/incentives/src/contract.rs
```diff
@@ -32,6 +32,10 @@ use crate::{
 pub const CONTRACT_NAME: &str = "crates.io:mars-incentives";
 pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
+/// The epoch duration should be at least one week, perhaps ideally one month. This is to ensure
+/// that the max gas limit is not reached when iterating over incentives.
+pub const MIN_EPOCH_DURATION: u64 = 604800u64;
```
grod220, line 37: Should this be hardcoded or a config variable?

grod220 (approved):

contracts/incentives/src/contract.rs
dancreee: I think this is a hardcoded value to enforce that the epoch_duration that is passed during contract instantiation is at least >= 1 week, to mitigate issues with dos type attacks against the contract, so there is no circumstance we want this contract instantiated with less than 1 week epochs

dancreee (approved):

contracts/incentives/src/contract.rs
dancreee: A bit "dirty" hardcoding this I guess but it's a bit of a limitation of the design we went for allowing n number of incentives against a market with the potential to get out of gas errors on a tx if someone manages to add many incentives to a market

apollo-sturdy: Yeah that's right. Epoch duration should never change as it would break the current incentives. If epoch duration is too short then there could be gas issues as we have to iterate over passed epochs when updating indexes. Of course we are the only ones who will be instantiating the contract (unless someone forks it), so we can just set it to a reasonable amount, but this hardcoded limit is just so we don't forget that fact.

contracts/incentives/src/contract.rs
```diff
@@ -57,6 +61,12 @@ pub fn instantiate(
     };
     CONFIG.save(deps.storage, &config)?;
 
+    if msg.epoch_duration < MIN_EPOCH_DURATION {
+        return Err(ContractError::EpochDurationTooShort {
+            min_epoch_duration: MIN_EPOCH_DURATION,
+        });
```
thec00n, line 67: Could we cover this with a test?

thec00n (commented):
LGTM

PR #223, opened 2023-06-26
fix: Error on too few snapshots and too short

apollo-sturdy:
window size


contracts/oracle/wasm/src/price_source.rs
```diff
@@ -174,6 +174,12 @@ impl PriceSourceUnchecked<WasmPriceSourceChecked, Empty> for WasmPriceSourceUnch
                     &route_assets,
                 )?;
 
+                if window_size <= 1 {
```
pacmanifold, line 177: set this either as const `MIN_WINDOW_SIZE` or config item

contracts/oracle/wasm/src/price_source.rs
```diff
@@ -307,6 +313,10 @@ fn query_astroport_twap_price(
         .may_load(deps.storage, denom)?
         .ok_or(ContractError::NoSnapshots {})?;
 
+    if snapshots.len() < 2 {
```
pacmanifold, line 316: set as either const `MIN_NUM_OF_SNAPSHOTS` or config item

dancreee (approved):

contracts/oracle/wasm/src/price_source.rs
apollo-sturdy: Could set it as a const if you prefer, but it shouldn't be changeable. Its just that the logic breaks (subtraction underflow) if you set it to 1 or less.

grod220 (approved):

PR #229, opened 2023-07-03
fix: Error on duplicate denoms in update whitelist args

apollo-sturdy:
Solves Oak audit issue 8

### DIFF
--- a/contracts/incentives/src/contract.rs
+++ b/contracts/incentives/src/contract.rs
@@ -148,8 +148,7 @@ pub fn execute_update_whitelist(
     let config = CONFIG.load(deps.storage)?;
 
     // Add add_denoms and remove_denoms to a set to check for duplicates
-    let denoms =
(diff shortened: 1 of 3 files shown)


contracts/incentives/src/contract.rs
```diff
@@ -147,6 +147,18 @@ pub fn execute_update_whitelist(
 
     let config = CONFIG.load(deps.storage)?;
 
+    // Add add_denoms and remove_denoms to a set to check for duplicates
+    let denoms =
+        add_denoms.iter().map(|(denom, _)| denom).chain(remove_denoms.iter()).collect::<Vec<_>>();
```
piobab, line 152: You don't have to `collect`. This way you increase num of operations. Just pass iterator to loop
```
let denoms =
        add_denoms.iter().map(|(denom, _)| denom).chain(remove_denoms.iter());
```

pacmanifold (approved):

piobab (approved):

PR #230, opened 2023-07-04
Remove deposit "on behalf of" [audit]

grod220:
Currently, when a new asset is listed, someone can deposit on behalf of Credit Manager. If they do this before any other lends have occurred there, they can cause it to throw errors for subsequent lenders via Credit Manager.

The reason is, Credit Manager is not expecting someone to be able to deposit to Red Bank on behalf of itself. After discussing with Piotr and others, we cannot think of a good reason why this functionality exists.

This PR removes `on_behalf_of` field from Red Bank, solving the Credit Manager vulnerability at the same time.

**Note:**: Running fmt/clippy found a few things, hence some unrelated code fixes/formatting.

### DIFF
--- a/contracts/oracle/base/src/pyth.rs
+++ b/contracts/oracle/base/src/pyth.rs
@@ -148,7 +148,8 @@ pub fn assert_pyth_current_price_not_too_old(
         return Err(InvalidPrice {
             reason: format!(
                 "current price publish time is too old/stale. published: {}, now: {}",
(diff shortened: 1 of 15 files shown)


piobab (approved):
LGTM!

`on_behalf_of` for deposit was for normal users too but I think it is not used in this case too.

thec00n (approved):

PR #235, opened 2023-07-05
[Oak Neutron Audit] Issue 10

pacmanifold:
Validate that Astroport TWAP window size is larger than tolerance

### DIFF
--- a/contracts/oracle/wasm/src/price_source.rs
+++ b/contracts/oracle/wasm/src/price_source.rs
@@ -6,7 +6,7 @@ use astroport::{
     querier::{query_token_precision, simulate},
 };
 use cosmwasm_schema::cw_serde;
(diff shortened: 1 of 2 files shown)


contracts/oracle/wasm/src/price_source.rs
```diff
@@ -165,6 +165,12 @@ impl PriceSourceUnchecked<WasmPriceSourceChecked, Empty> for WasmPriceSourceUnch
                 tolerance,
                 route_assets,
             } => {
+                if tolerance >= window_size {
+                    return Err(
+                        StdError::generic_err("tolerance must be less than window size").into()
```
piobab, line 170: Could you use `InvalidPriceSource` error type?

piobab (approved):

piobab (approved):

PR #238, opened 2023-07-05
Shared deposit cap

larry0x:


larry0x:
Closing this PR because we have decided on an alternative approach (putting the logics in mars-params instead) and the commit history of this PR is pretty messed up

PR #239, opened 2023-07-06
Fix/oak neutron audit issue 3

piobab:


contracts/red-bank/src/execute.rs
```diff
@@ -422,6 +420,9 @@ pub fn deposit(
     )?;
 
     market.increase_collateral(deposit_amount_scaled)?;
+
+    response = update_interest_rates(&env, &mut market, response)?;
+
```
apollo-sturdy, line 425: Can you explain how this change relates to the audit issue?

apollo-sturdy (approved):
Another suggestion for defensive programming could be to limit `utilization_rate` to 100% when calculating it. With the current code it should hopefully never happen that it gets calculated to more than 100%, but I also don't see any downside to doing e.g. `let utilization_rate = max(utilization_rate, Decimal::percent(100))`. In the case where we have missed something this might save us from the exploit.

contracts/red-bank/src/execute.rs
piobab: It corresponds to the case that once we moved everything to internally tracked balances we should also correctly update interest rates based on the changes to collaterals (see the lines above)

PR #241, opened 2023-07-06
[Oak Neutron Audit] Issue 2

pacmanifold:
Removes route_assets from WasmPriceSource. Since all of our price sources report the price in the base denom, we don't need route assets. All we need to make sure is that when adding a price source that uses an astroport pair, that does not contain the base denom, there must already exist a price source for the other denom of that pool.

Also removes two test cases. Reasoning is that if "uosmo" is the base denom the prices should be reported in "uosmo" and having a route asset ("uosmo" = 2) does not make sense.


contracts/oracle/wasm/src/helpers.rs
```diff
@@ -37,74 +35,59 @@ pub fn astro_native_asset(denom: impl Into<String>, amount: impl Into<Uint128>)
...
-        // If there are no route assets, then the pair must contain the denom and base denom.
-        assert_astroport_pair_contains_denoms(&pair_info, &[denom, base_denom])?;
-    } else {
-        // If there are route assets, the pair must contain the denom and the first route asset.
-        // The rest should already be validated in the same way because we checked above that a
-        // price source exists for each of them, and the corresponding validation would have been
-        // done when they were added.
-        assert_astroport_pair_contains_denoms(&pair_info, &[denom, &route_assets[0]])?;
+    // Pair must contain two assets
+    if pair_denoms.len() != 2 {
```
piobab, line 66: Move it up before you start loading data from state. Put it just after `let pair_denoms = get_astroport_pair_denoms(deps, pair_address)?;`

contracts/oracle/wasm/src/helpers.rs
```diff
@@ -140,57 +123,34 @@ pub fn period_diff(
...
 ) -> ContractResult<Decimal> {
-    let mut price = *price;
-    for denom in route_assets {
-        let price_source = price_sources.load(deps.storage, denom)?;
-        let route_price = price_source.query_price(deps, env, denom, config, price_sources)?;
-        price *= route_price;
+    let mut pair_denoms = get_astroport_pair_denoms(deps, pair_address)?;
+
+    if pair_denoms.contains(&config.base_denom) {
+        deps.api.debug(&format!("Pair {} contains base denom {}", pair_address, config.base_denom));
```
piobab, line 141: To be removed?

contracts/oracle/wasm/src/helpers.rs
```diff
@@ -140,57 +123,34 @@ pub fn period_diff(
...
+        Ok(price)
+    } else {
+        pair_denoms.retain(|d| d != denom);
```
piobab, line 144: Maybe add comment if someone doesn't check validate that we assert that there are only 2 denoms in the pair.

piobab (commented):
LGTM, just small comments. You have to fix build too.

contracts/oracle/wasm/src/helpers.rs
pacmanifold: fixed

contracts/oracle/wasm/src/helpers.rs
pacmanifold: fixed

contracts/oracle/wasm/src/helpers.rs
pacmanifold: done

piobab (approved):

contracts/oracle/wasm/tests/test_price_source.rs
```diff
@@ -196,7 +186,6 @@ pub fn test_validate_and_query_astroport_spot_price_source(
 #[test_case(PairType::Xyk {}, &["uatom","uosmo"], "uosmo", &[], 5, 100; "XYK, no route, base_denom in pair")]
 #[test_case(PairType::Xyk {}, &["uatom","uosmo"], "USD", &[], 5, 100 => panics; "XYK, no route, base_denom not in pair")]
 #[test_case(PairType::Xyk {}, &["uatom","uosmo"], "uosmo", &[("uusd", TWO)], 5, 100 => panics; "XYK, route asset does not exist")]
-#[test_case(PairType::Xyk {}, &["uatom","uosmo"], "uosmo", &[("uosmo", TWO)], 5, 100; "XYK, route equal to base_denom")]
 #[test_case(PairType::Xyk {}, &["uatom","uosmo"], "uosmo", &[("uion",TWO)], 5, 100 => panics; "XYK, route with non-base existing asset, not in pair")]
 #[test_case(PairType::Xyk {}, &["uatom","uion"], "uosmo", &[("uion",TWO)], 5, 100; "XYK, route with non-base existing asset, in pair")]
 #[test_case(PairType::Stable {}, &["uatom","uosmo"], "uosmo", &[], 5, 100; "Stable, no route, base_denom in pair")]
```
apollo-sturdy, line 191: Remove route_prices argument?

apollo-sturdy (changes requested):

apollo-sturdy (approved):
Fixed some issues. LGTM now but should review my changes too.

piobab (approved):

contracts/oracle/wasm/src/helpers.rs
```diff
@@ -37,76 +35,69 @@ pub fn astro_native_asset(denom: impl Into<String>, amount: impl Into<Uint128>)
...
+            AssetInfo::Token {
+                contract_addr,
+            } => Err(ContractError::InvalidPriceSource {
+                reason: format!("pair contains cw20 token: {}", contract_addr),
+            }),
+            AssetInfo::NativeToken {
+                denom,
+            } => Ok(denom.clone()),
+        })
+        .collect()
```
pacmanifold, line 98: break this out into separate function so it can be unit tested?

packages/testing/src/wasm_oracle.rs
```diff
@@ -386,42 +394,44 @@ pub fn validate_and_query_astroport_spot_price_source(
...
         .add_denom_precision_to_coin_registry(base_denom, 6, admin)
-        .set_price_sources(route_price_sources, admin)
-        .set_price_source(pair_denoms[0], price_source.clone(), admin)
-        .assert_price_source(pair_denoms[0], price_source)
-        .assert_price_almost_equal(pair_denoms[0], expected_price, Decimal::permille(1));
+        .set_price_sources(other_asset_price_source, admin);
+
+    robot.set_price_source(primary_denom, price_source.clone(), admin);
+
+    robot.assert_price_source(primary_denom, price_source).assert_price_almost_equal(
```
pacmanifold, line 421: why not chain on previous call?

contracts/oracle/wasm/src/helpers.rs
apollo-sturdy: Will do in another PR

packages/testing/src/wasm_oracle.rs
apollo-sturdy: Will do in another PR

contracts/oracle/wasm/src/helpers.rs
apollo-sturdy: https://github.com/mars-protocol/red-bank/pull/243

packages/testing/src/wasm_oracle.rs
apollo-sturdy: https://github.com/mars-protocol/red-bank/pull/243

PR #242, opened 2023-07-07
Update interest rate test case.

piobab:


piobab:
@pacmanifold @apollo-sturdy would like to merge but would be nice to have one review

apollo-sturdy (approved):
lgtm. Maybe consider commiting the spreadsheet to the repo so we don't rely on google?

PR #246, opened 2023-07-13
fix: Fix range bounds in QueryMsg::Emission

apollo-sturdy:


piobab (approved):

pacmanifold:
@apollo-sturdy This was the fix i though should work. How ever i tried querying the current contract with `current_time - epoch_duration` and it still returned 0. Only when providing exactly the right timestamp did it return non-zero

Have you tried to see if this works?

contracts/incentives/tests/test_quering.rs
```diff
@@ -143,6 +143,28 @@ fn query_emission() {
...
+    // Query at timestamp some time after emission end
+    let res: Uint128 = helpers::th_query(
+        deps.as_ref(),
+        QueryMsg::Emission {
+            collateral_denom: "uatom".to_string(),
+            incentive_denom: "umars".to_string(),
+            timestamp: 604800 * 3,
+        },
+    );
+    assert_eq!(res, Uint128::zero());
```
pacmanifold, line 167: lets add one test case querying before emission start

pacmanifold (approved):
one comment. otherwise lgtm

pacmanifold (approved):
lgtm

PR #265, opened 2023-07-20
Disable grcov coverage

piobab:
Disable grcov coverage. It blocks CI with `no space left` or `timeout`. This requires more investigation so for now disabling coverage in CI.


Makefile.toml
```diff
@@ -21,6 +21,7 @@ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profil
 rustup target add wasm32-unknown-unknown --toolchain ${RUST_VERSION}
 rustup component add rustfmt --toolchain ${RUST_VERSION}
 rustup component add clippy --toolchain ${RUST_VERSION}
+rustup component add llvm-tools-preview --toolchain ${RUST_VERSION}
```
apollo-sturdy, line 24: Is it needed?

apollo-sturdy (approved):

Makefile.toml
piobab: yeap, it was before and I accidentally removed it. See `usage` section https://crates.io/crates/grcov/0.8.19

PR #270, opened 2023-07-24
New dynamic liquidation mechanics.

piobab:
- liquidation with simplified protocol fee as part of total liquidation bonus
- improved test cases (used multi-test)
- move liquidate to separate file (in next PR I will do the same for deposit, withdraw etc)


contracts/red-bank/src/liquidate.rs
```diff
@@ -0,0 +1,378 @@
...
+/// liq_th_collateral_value     - Value of collateral before the liquidation happens adjusted to liquidation threshold
+/// requested_collateral_liq_th - Liquidation threshold of requested collateral
+/// LB                          - Liquidation Bonus
+///
+/// PLF (Protocol Liqudiation Fee) is charged as a % of the LB.
+/// For example, if we define the PLF as 10%, then the PLF would be deducted from the LB, so upon a liquidation:
+/// - The liquidator receives 90% of the LB.
+/// - The remaining 10% is sent to the protocol as PLF.
+#[allow(clippy::too_many_arguments)]
+pub fn calculate_liquidation_amounts(
```
piobab, line 257: I am thinking to move it to separate package to be reused between red bank and rover. Only question is test coverage because red-bank and rover have different one.

brimigs (commented):
We'll need to update all the types in the deploy scripts but we can just add that to the bulleted list in the upcoming build scripts jira ticket

brimigs (commented):
It also looks like we'll need to update the github actions for running tests: https://stackoverflow.com/questions/28124221/error-linking-with-cc-failed-exit-code-1

piobab:
> It also looks like we'll need to update the github actions for running tests: https://stackoverflow.com/questions/28124221/error-linking-with-cc-failed-exit-code-1

Problem is more complex:
```
note: /usr/bin/ld: final link failed: No space left on device
          collect2: error: ld returned 1 exit status
```

.github/workflows/coverage.yml
```diff
@@ -26,8 +26,8 @@ jobs:
         run: cargo make install-stable
 
       # selecting a toolchain should happen before the plugin, as the cache uses the current rustc version as its cache key
-      - name: Cache dependencies
-        uses: Swatinem/rust-cache@v2
+      # - name: Cache dependencies
+      #   uses: Swatinem/rust-cache@v2
```
grod220, line 30: Should we just remove this if it's broken?

contracts/red-bank/src/liquidate.rs
grod220: Think a separate package works. Rover's test coverage in theory should not change by adding that dependency and removing the code on that side.

contracts/red-bank/tests/files/Red Bank - Dynamic LB & CF test cases v1.1.xlsx
grod220, line 1: Is it better to have this versus a link?

scripts/deploy/osmosis/config.ts
```diff
@@ -26,109 +26,175 @@ const marsOsmoPool = 907
...
-  deposit_enabled: true,
-  borrow_enabled: true,
+  protocol_liquidation_fee: '0.5',
+  // reserve_factor: '0.2',
+  // interest_rate_model: {
+  //   optimal_utilization_rate: '0.6',
+  //   base: '0',
+  //   slope_1: '0.15',
+  //   slope_2: '3',
+  // },
```
grod220, line 43: We can delete the commented code in this file?

grod220 (commented):
I think it's worthwhile exporting to a package now versus later. What do you think?

.github/workflows/coverage.yml
```diff
@@ -14,7 +14,7 @@ env:
 
 jobs:
   coverage:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
```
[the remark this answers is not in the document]
piobab: Yeap, it was introduced on Sturdy Pr but anyway it doesn't help. And I agree with you.

contracts/red-bank/src/execute.rs
```diff
@@ -691,281 +685,6 @@ pub fn repay(
...
-    let refund_amount = sent_debt_amount - debt_amount_to_repay;
-
-    Ok((
-        debt_amount_to_repay,
-        collateral_amount_to_liquidate,
-        collateral_amount_to_liquidate_scaled,
-        refund_amount,
-    ))
-}
-
```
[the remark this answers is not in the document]
piobab: It is time to refactor this codebase ;)

.github/workflows/coverage.yml
piobab: Yeap, i will do. It is my testing to pass CI :/

contracts/red-bank/tests/files/Red Bank - Dynamic LB & CF test cases v1.1.xlsx
piobab: Delphi Labs docs are not available outside of the company. IMO, I think is much better to keep it versioned in codebase.

scripts/deploy/osmosis/config.ts
piobab: Yeap, in next PRs. Leaving it just for now for reference.

contracts/red-bank/src/liquidate.rs
piobab: Wdyt about adding this liquidation to health package? Then liquidate function can be implemented for Health struct.. Not sure what is best

grod220: The health package relates so strongly to credit manager though 🤔 . I kinda think liquidation on its own would be fine.

piobab: Hmm we have rover and redbank health packages but yeah make sense to create separate package

brimigs (approved):

contracts/red-bank/src/liquidate.rs
piobab: Extracted

.github/workflows/coverage.yml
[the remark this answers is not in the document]
piobab: Removed

.github/workflows/coverage.yml
piobab: Removed

grod220 (approved):
Shouldn't the tests live in the liquidation package? Not sure, but that's my only additional thought. Good job!

PR #273, opened 2023-07-25
MP-2615 rover rewards

piobab:
Add support to track rover account ids in deposit/withdraw

Note: All migration for red bank will go in separate PR.


contracts/incentives/src/state.rs
```diff
@@ -32,12 +32,12 @@ pub const INCENTIVE_STATES: Map<(&str, &str), IncentiveState> = Map::new("incent
 pub const EMISSIONS: Map<(&str, &str, u64), Uint128> = Map::new("emissions");
 
 /// A map containing the incentive index for a given user, collateral denom and incentive denom.
-/// The key is (user address, collateral denom, incentive denom).
-pub const USER_ASSET_INDICES: Map<(&Addr, &str, &str), Decimal> = Map::new("indices");
+/// The key is (user address or account id, collateral denom, incentive denom).
+pub const USER_ASSET_INDICES: Map<(&str, &str, &str), Decimal> = Map::new("indices");
```
piobab, line 36: Hmm maybe better to keep (&Addr, &str) even duplicating `Addr` for rover accounts. It can be consistent with redbank.

piobab: Added in another commit. Will revert in case not accepted by others

contracts/incentives/src/state.rs
```diff
@@ -32,12 +32,13 @@ pub const INCENTIVE_STATES: Map<(&str, &str), IncentiveState> = Map::new("incent
 pub const EMISSIONS: Map<(&str, &str, u64), Uint128> = Map::new("emissions");
 
 /// A map containing the incentive index for a given user, collateral denom and incentive denom.
-/// The key is (user address, collateral denom, incentive denom).
-pub const USER_ASSET_INDICES: Map<(&Addr, &str, &str), Decimal> = Map::new("indices");
+/// The key is (user address or account id, collateral denom, incentive denom).
```
grod220, line 35: isn't it `user address with optional account id`

contracts/incentives/tests/test_balance_change.rs
```diff
@@ -278,9 +284,16 @@ fn with_zero_previous_balance_and_asset_with_zero_index_accumulates_rewards() {
...
-            query_user_unclaimed_rewards(deps.as_ref(), env, "user".to_string(), None, None, None)
-                .unwrap();
+        let rewards_query = query_user_unclaimed_rewards(
+            deps.as_ref(),
+            env,
+            "user".to_string(),
+            None,
+            None,
+            None,
+            None,
```
grod220, line 294: Yeesh, so many none's. Wish rust had named arguments lol.

contracts/incentives/tests/test_balance_change.rs
grod220, line 1: Do we need an account_id test here? They all seem none.

contracts/incentives/tests/test_claim_rewards.rs
grod220, line 1: An account_id test here too?

contracts/red-bank/src/collateral.rs
```diff
@@ -21,18 +21,17 @@ pub fn update_asset_collateral_status(
...
+        .ok_or_else(|| ContractError::UserNoCollateralBalance {
+            user: user.into(),
+            denom: denom.clone(),
         })?;
 
     let previously_enabled = collateral.enabled;
 
     collateral.enabled = enable;
-    COLLATERALS.save(deps.storage, (user.address(), &denom), &collateral)?;
+    COLLATERALS.save(deps.storage, (user.address(), "", &denom), &collateral)?;
```
grod220, line 34: Wouldn't this still allow them to use the ones that are associated with account ids?

contracts/red-bank/src/health.rs
```diff
@@ -117,7 +117,7 @@ pub fn get_user_positions_map(
 
     // Find all denoms that the user has a collateral or debt position in
     let collateral_denoms = COLLATERALS
-        .prefix(user_addr)
+        .prefix((user_addr, ""))
```
grod220, line 120: We shouldn't assume none here right?

grod220 (commented):
There are usages in the contracts where account ID is assumed to be `""`. I wonder if that poses safety issues. Shouldn't all of those function signatures accept `account_id` now?

contracts/red-bank/src/collateral.rs
piobab: We don't support enabling / disabling collateral for rover accounts so I think no. By default collateral for rover accounts are enabled.

contracts/red-bank/src/health.rs
piobab: hmm maybe for safety I will add account_id: Option<String>. We handle health in different way for redbank and rover so basically this function is used by redbank users.

contracts/incentives/tests/test_claim_rewards.rs
piobab: see `rewards_claim_for_credit_account`.

contracts/incentives/tests/test_balance_change.rs
piobab: Yeap, I wish too.

grod220 (approved):

brimigs (approved):

PR #279, opened 2023-07-27
Migrate to PoolManager for Pool and SpotPrice. Read StableSwap pool.

piobab:
- migrate Pool and SpotPrice request to PoolManager
- read/parse StableSwap pool

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ neutron-sdk       = "0.6.0"
 # dev-dependencies
 cw-multi-test     = "0.16.5"
 cw-it             = "0.1.0"
-osmosis-test-tube = "16.0.0"
+osmosis-test-tube = "=16.1.1"
 test-case         = "3.0.0"
 proptest          = "1.1.0"
 
--- a/contracts/oracle/osmosis/src/helpers.rs
+++ b/contracts/oracle/osmosis/src/helpers.rs
@@ -1,28 +1,63 @@
 use mars_oracle_base::{ContractError, ContractResult};
-use mars_osmosis::helpers::{has_denom, Pool};
+use mars_osmosis::{
+    helpers::{CommonPoolData, Pool},
+    BalancerPool,
+};
 
 use crate::DowntimeDetector;
 
 /// 48 hours in seconds
 const TWO_DAYS_IN_SECONDS: u64 = 172800u64;
 
-/// Assert the Osmosis pool indicated by `pool_id` is of XYK type and assets are OSMO and `denom`
+/// Assert the Osmosis pool indicated by `pool_id` is of Balancer XYK or StableSwap and assets are OSMO and `denom`
 pub fn assert_osmosis_pool_assets(
     pool: &Pool,
     denom: &str,
     base_denom: &str,
 ) -> ContractResult<()> {
-    assert_osmosis_xyk_pool(pool)?;
+    match pool {
+        Pool::Balancer(balancer_pool) => {
+            assert_osmosis_xyk_pool(balancer_pool)?;
+        }
+        Pool::StableSwap(_) => {}
+    };
 
-    if !has_denom(base_denom, &pool.pool_assets) {
+    assert_osmosis_pool_contains_two_assets(pool, denom, base_denom)?;
+
+    Ok(())
+}
+
+/// Assert the Osmosis pool indicated by `pool_id` is Balancer XYK type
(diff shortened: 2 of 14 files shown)


brimigs (approved):

Cargo.toml
```diff
@@ -62,7 +62,7 @@ neutron-sdk       = "0.6.0"
 # dev-dependencies
 cw-multi-test     = "0.16.5"
 cw-it             = "0.1.0"
-osmosis-test-tube = "16.0.0"
+osmosis-test-tube = "=16.1.1"
```
grod220, line 65: Yeah, it's probably best if we pin these versions. This library is quite unstable 😅

contracts/oracle/osmosis/src/helpers.rs
```diff
@@ -1,36 +1,71 @@
...
+
+/// Assert the Osmosis pool indicated by `pool_id` is Balancer XYK type
+pub fn assert_osmosis_xyk_lp_pool(pool: &Pool) -> ContractResult<()> {
+    match pool {
+        Pool::Balancer(balancer_pool) => assert_osmosis_xyk_pool(balancer_pool)?,
+        Pool::StableSwap(stable_swap_pool) => {
+            return Err(ContractError::InvalidPriceSource {
+                reason: format!("StableSwap pool not supported. Pool id {}", stable_swap_pool.id),
+            });
+        }
```
grod220, line 38: I'm confused why you need a match here if the above already filtered for Balancer pools. Shouldn't this be like:

```rust
pub fn assert_osmosis_xyk_lp_pool(pool: &BalancerPool)
```

packages/chains/osmosis/src/helpers.rs
```diff
@@ -3,36 +3,74 @@ use std::str::FromStr;
...
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
-pub struct Pool {
-    pub id: String,
-    pub address: String,
-    pub pool_params: Option<PoolParams>,
-    pub future_pool_governor: String,
-    pub pool_assets: Vec<PoolAsset>,
-    pub total_shares: Option<Coin>,
-    pub total_weight: String,
+use prost::Message;
```
grod220, line 21: Is this used anywhere?

Cargo.toml
piobab: Yeap, i had different versions resolved and felt a pain :) Leave it like this.

contracts/oracle/osmosis/src/helpers.rs
piobab: Maybe I don't understand correctly `if the above already filtered for Balancer pools`...

BUT

We would like to allow to setup Balancer XYK pool (equal weights) and StableSwap for most of the price sources on Osmosis (in reality only one price source will use StableSwap - `StakedGeometricTwap`). For `XYK_Liquidity_Token` we don't have pricing methodology yet so we have to assert Balancer XYK pool.

packages/chains/osmosis/src/helpers.rs
piobab: Yes, it is used for `decode`

contracts/oracle/osmosis/src/helpers.rs
grod220: In this section:

```rust
    match pool {
        Pool::Balancer(balancer_pool) => {
            assert_osmosis_xyk_pool(balancer_pool)?;
        }
        Pool::StableSwap(_) => {}
    };
```
`assert_osmosis_xyk_pool` should have it filtered down to only balancer pools at this point. Was confused by the match inside the function. But maybe I don't understand how it works exactly.

grod220 (approved):

contracts/oracle/osmosis/src/helpers.rs
piobab: `assert_osmosis_xyk_pool` (assert if it is Balancer pool) is different than `assert_osmosis_xyk_lp_pool` (we pass all pools here and check if it is Balancer XYK pool) -> maybe this confused you.

piobab: Basically we can't use in `XYK_liquidity_token` validation because we can't allow StableSwap here (second branch in the match pass further):
```
match pool {
        Pool::Balancer(balancer_pool) => {
            assert_osmosis_xyk_pool(balancer_pool)?;
        }
        Pool::StableSwap(_) => {}
    };
```

PR #280, opened 2023-07-28
Shared deposit cap

larry0x:
1. Extract interest rate-related logics in `mars-red-bank` to a `mars-interest-rate` library
3. Make Red Bank's `Market` query to return `Option<Market>`; `None` if there no market is found for the given denom
4. Add a `TotalDeposit` query to `mars-params` that computes the total amount of a denom deposited across Red Bank and Rover. NOTE: we only account for spot holding; we don't consider LP tokens or vaults, which we have concluded as infeasible after discussions.
5. Modify Red Bank's `deposit` function to assert deposit cap is not exceeded.


contracts/red-bank/src/deposit.rs
```diff
@@ -45,10 +43,8 @@ pub fn deposit(
         });
     }
 
-    let total_scaled_deposits = market.collateral_total_scaled;
-    let total_deposits =
-        get_underlying_liquidity_amount(total_scaled_deposits, &market, env.block.time.seconds())?;
-    if total_deposits.checked_add(deposit_amount)? > asset_params.red_bank.deposit_cap {
+    let total_deposits = query_total_deposit(&deps.querier, params_addr, &denom)?;
+    if total_deposits.amount.checked_add(deposit_amount)? > asset_params.red_bank.deposit_cap {
```
larry0x, line 47: here's the most important change. instead of checking the `Market` stored inside red bank, we query the params contract for the total deposit amount

contracts/params/src/query.rs
```diff
@@ -49,3 +54,78 @@ pub fn query_all_vault_configs(
...
+                user: credit_manager_addr.into(),
+                denom: denom.clone(),
+            },
+        )?
+        .amount;
+
+    // amount of this asset deposited into Credit Manager
+    // this is simply the coin balance of the CM contract
+    // note that this way, we don't include LP tokens or vault positions
+    let cm_deposit = deps.querier.query_balance(credit_manager_addr, &denom)?.amount;
```
piobab, line 125: Do you think we are ok if someone send tokens directly? I mean it is not a big problem but can influence internally tracked balances.

contracts/red-bank/src/execute.rs
```diff
@@ -0,0 +1,1019 @@
+use std::{cmp::min, str};
```
piobab, line 1: Could you verify if you are rebased? It is splitted to different files.

contracts/params/src/query.rs
larry0x: Good point. Given this implementation, an attacker can prevent anyone else from making new deposits, since by sending large quantity of coins to the CM contract, tricking mars-params into thinking the cap has been reached. 

However I can't think of anyone the attacker can profit from this, and we can easily fix the problem and confiscate his funds via a simple contract migration. Therefore I think the likelihood of this attack is low.

I'm not sure whether CM tracks the total deposit, if not then we need to add it.

contracts/red-bank/src/execute.rs
piobab: This file should be removed.

piobab (approved):
LGTM. Pushed fixes for tests and build.

brimigs (approved):

piobab:
Moved `deposit_cap` outside of `RedBankSettings`.

PR #281, opened 2023-07-30
Use `cosmwasm_std::Coins` in the accounting of unclaimed rewards

larry0x:
this avoids generating a `BankMsg::Send` that contains coins with zero amounts.

### DIFF
--- a/contracts/incentives/src/contract.rs
+++ b/contracts/incentives/src/contract.rs
@@ -1,10 +1,8 @@
-use std::collections::HashMap;
-
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
-    attr, coins, to_binary, Addr, BankMsg, Binary, Coin, CosmosMsg, Decimal, Deps, DepsMut, Env,
-    Event, MessageInfo, Order, Response, StdError, StdResult, Uint128,
+    attr, to_binary, Addr, BankMsg, Binary, Coin, Coins, Decimal, Deps, DepsMut, Env, Event,
+    MessageInfo, Order, Response, StdError, StdResult, Uint128,
 };
 use cw_storage_plus::Bound;
 use mars_owner::{OwnerInit::SetInitialOwner, OwnerUpdate};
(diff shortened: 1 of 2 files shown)


contracts/incentives/src/contract.rs
```diff
@@ -456,22 +454,27 @@ pub fn execute_claim_rewards(
...
+    if !total_unclaimed_rewards.is_empty() {
+        let coins: Vec<Coin> = total_unclaimed_rewards.into();
+
+        for coin in &coins {
+            events.push(
+                Event::new("mars/incentives/claim_rewards/claimed_reward")
+                    .add_attribute("denom", coin.denom.clone())
+                    .add_attribute("amount", coin.amount),
+            );
+        }
```
larry0x, line 472: this section of the code is a bit verbose because `Coins` currently doesn't implement an iterator trait, so we have to cast it to a `Vec<Coin>` first.

i have created an issue on this: https://github.com/CosmWasm/cosmwasm/issues/1804. once fixed, we can simplify this section.

alternatively, instead of a separate event for each coin, we can have a single event for all reward coins:

```rust
let event = Event::new("mars/incentives/claim_rewards/claimed_rewards")
    .add_attribute("coins", total_unclaimed_rewards.to_string());

Ok(Response::new().add_event(event))
```

but we need backend/indexer team to agree on this. @piobab thoughts?

contracts/incentives/src/contract.rs
```diff
@@ -649,7 +652,7 @@ pub fn query_user_unclaimed_rewards(
         limit,
     )?;
 
-    let mut total_unclaimed_rewards: HashMap<String, Uint128> = HashMap::new();
+    let mut total_unclaimed_rewards = Coins::default();
 
     for ((collateral_denom, incentive_denom), _) in incentive_states {
```
larry0x, line 657: this section of code exists in both the execute and query functions. we should extract it to a separate helper function

larry0x: actually no, nevermind - in the execute function it needs to update the storage

dancreee (approved):

contracts/incentives/src/contract.rs
```diff
@@ -456,25 +454,25 @@ pub fn execute_claim_rewards(
...
-        events.push(
-            Event::new("mars/incentives/claim_rewards/claimed_reward")
-                .add_attribute("denom", denom)
-                .add_attribute("amount", *amount),
-        );
+    if !total_unclaimed_rewards.is_empty() {
+        response = response
+            .add_event(
+                Event::new("mars/incentives/claim_rewards/claimed_rewards")
+                    .add_attribute("amount", total_unclaimed_rewards.to_string()),
```
piobab, line 467: Maybe instead of `amount` -> `coins` ?

piobab (approved):

piobab:
@larry0x one comment. We can't squash/rebase merge. Please squash everything to one commit and the just `MERGE` so we can then do `git merge develop` -> `solve connflicts` and merge to develop branch.

contracts/incentives/src/contract.rs
larry0x: using `coins` now

PR #296, opened 2023-08-05
Mp 2830 withdraw protocol fee from rover

piobab:


contracts/rewards-collector/osmosis/tests/test_withdraw.rs
```diff
@@ -37,3 +41,76 @@ fn withdrawing_from_red_bank() {
         }))
     )
 }
+
+#[test]
+fn withdrawing_from_cm_if_action_not_allowed() {
```
piobab, line 46: It would be nice to test with rover but that requires copying or sharing all helpers from rover multitest. I think it's an option for improvement.

I assume that the rover is well covered with tests that it is enough to check the correct actions that go to it.

larry0x (approved):

PR #300, opened 2023-08-07
Clean up

larry0x:
- remove unused dependencies and features
- format `Cargo.toml`s (align items and sort alphabetically)
- update all dependencies to latest (by running `cargo update`)
- disable doctest for all crates (we don't need it, so disabling speeds up CI by a little bit)

### DIFF
--- a/contracts/address-provider/Cargo.toml
+++ b/contracts/address-provider/Cargo.toml
@@ -11,22 +11,13 @@ documentation = { workspace = true }
 keywords      = { workspace = true }
 
 [lib]
-crate-type = [
-    "cdylib",
-    "rlib",
-]
-doctest = false
-
-[profile.release]
-overflow-checks  = true
+crate-type = ["cdylib", "rlib"]
+doctest    = false
 
 [features]
 # for more explicit tests, cargo test --features=backtraces
-backtraces = [
-    "cosmwasm-std/backtraces",
-]
-library = [
-]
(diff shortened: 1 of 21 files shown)


contracts/address-provider/Cargo.toml
```diff
@@ -11,22 +11,13 @@ documentation = { workspace = true }
...
 
 [lib]
-crate-type = [
-    "cdylib",
-    "rlib",
-]
-doctest = false
-
-[profile.release]
-overflow-checks  = true
```
larry0x, line 21: Removed all instances of this, because we already configure this in the workspace `Cargo.toml`

contracts/swapper/base/Cargo.toml
```diff
@@ -10,10 +10,7 @@ documentation = { workspace = true }
 keywords      = { workspace = true }
 
 [lib]
-crate-type = ["cdylib", "rlib"]
```
larry0x, line 13: The `swapper-base` contract doesn't need to be compiled into a wasm binary, so we don't need this

brimigs (approved):

contracts/address-provider/Cargo.toml
piobab: Make sense. Oak usually points this in the reports.

piobab (approved):

PR #304, opened 2023-08-18
Validate if slippage not too high.

piobab:
The slippage parameter is not explicitly sanity checked. The parameter could be set 100% in which case the user could receive no output coins as a result.

*Remediation*
The slippage parameter should be checked if it’s less than 1 and not exceed maximum amount e.g. max slippage 10%.


brimigs (approved):
nice 👍

thec00n:
Introducing max slippage could be problematic when a pool has low liquidity. I think we should assert that the upstream function sets a max slippage but not limit at a certain percentage.

piobab:
> function sets a max slippage but not limit at a certain percentage.

Wdym?

I think we don't want to work on low liquidity pool.

thec00n:
It's fine as long as pools have enough liquidity. I just wanted to note that it can become a problem with low liquidity pools.

piobab:
@thec00n yeap, for sure but I think low liquidity pools are not our goal.

PR #306, opened 2023-08-24
MP-3292. Oak no. 1

piobab:
Lendings in the red bank cannot be liquidated when ActionKind::Default circuit breakers trigger


brimigs (approved):
LGTM

PR #312, opened 2023-08-29
MP-3296 pagination

piobab:
Common way to do pagination in Cosmwasm is using `limit`, `start_after` and respond with vector of data. For example:
```
// settings for pagination
const MAX_LIMIT: u32 = 30;
const DEFAULT_LIMIT: u32 = 10;

pub fn query_list_members(
    deps: Deps,
    start_after: Option<String>,
    limit: Option<u32>,
) -> StdResult<MemberListResponse> {
    let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
    let addr = maybe_addr(deps.api, start_after)?;
    let start = addr.as_ref().map(Bound::exclusive);

    let members = MEMBERS
        .range(deps.storage, start, None, Order::Ascending)
        .take(limit)
        .map(|item| {
            item.map(|(addr, weight)| Member {
                addr: addr.into(),
                weight,
            })
        })
        .collect::<StdResult<_>>()?;

    Ok(MemberListResponse { members })
}
```

As an API user if I want to know if there is more data:
- check if responded number of elements is less than sent `limit` parameter. This is problematic because if I send `limit` > `MAX_LIMIT` I get less elements than requested - thinking that there is no more data.
- execute next query and see if there is more data. This is not efficient because I make extra query.
Moreover, it is easy to forget about pagination as we noticed during contract configuration / validation. You could query something and you basically use default pagination params and don't see the rest of elements.

*Solution*
Add metadata data
```
#[cw_serde]
pub struct PaginationResponse<T> {
    pub data: Vec<T>,
    pub metadata: Metadata,
}

#[cw_serde]
pub struct Metadata {
    pub has_more: bool,
    pub next_start_after: Option<String>,
}
```
This way we see in the response if there is more data. Easy to operate on without depending on default params.

### DIFF
--- a/packages/types/src/lib.rs
+++ b/packages/types/src/lib.rs
@@ -5,3 +5,17 @@ pub mod oracle;
 pub mod red_bank;
 pub mod rewards_collector;
 pub mod swapper;
+
+use cosmwasm_schema::cw_serde;
+
+#[cw_serde]
+pub struct PaginationResponse<T> {
+    pub data: Vec<T>,
+    pub metadata: Metadata,
+}
+
+#[cw_serde]
+pub struct Metadata {
+    pub has_more: bool,
+    pub next_start_after: Option<String>,
+}
--- a/packages/types/src/lib.rs
+++ b/packages/types/src/lib.rs
@@ -17,5 +17,4 @@ pub struct PaginationResponse<T> {
 #[cw_serde]
 pub struct Metadata {
     pub has_more: bool,
-    pub next_start_after: Option<String>,
 }
--- a/packages/types/src/red_bank/msg.rs
+++ b/packages/types/src/red_bank/msg.rs
@@ -195,6 +195,15 @@ pub enum QueryMsg {
         limit: Option<u32>,
     },
 
+    /// Get all collateral positions for a user
+    #[returns(Vec<crate::red_bank::PaginatedUserCollateralResponse>)]
+    UserCollateralsV2 {
+        user: String,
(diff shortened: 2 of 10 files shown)


packages/types/src/lib.rs
```diff
@@ -5,3 +5,17 @@ pub mod oracle;
...
+#[cw_serde]
+pub struct PaginationResponse<T> {
+    pub data: Vec<T>,
+    pub metadata: Metadata,
+}
+
+#[cw_serde]
+pub struct Metadata {
+    pub has_more: bool,
+    pub next_start_after: Option<String>,
```
larry0x, line 20: I think `next_start_after` is redundant, because the querier can simply take the last element in the `Vec`. This just increases the response data size.

packages/types/src/red_bank/msg.rs
```diff
@@ -195,6 +195,15 @@ pub enum QueryMsg {
         limit: Option<u32>,
     },
 
+    /// Get all collateral positions for a user
+    #[returns(Vec<crate::red_bank::PaginatedUserCollateralResponse>)]
```
larry0x, line 199: I think this should be just `PaginatedUserCollateralResponse`, not need to the `Vec`

piobab: Ahh yes yes. Copy paste bug

thec00n (approved):

dancreee (approved):

PR #314, opened 2023-08-31
MP-3299, Oak no. 8, contract version update. MP-3310, Oak no. 19, provide whitelist count.

brimigs:
The incentive contract inherits the version of the workspace in red-bank:Cargo.toml. During initialization, the version is indeed set correctly. However, during migration, the contract version is not updated in red-bank:contracts/incentives/src/contract.rs:767-769.
This could lead to confusion during future migrations as it will not be possible to be certain which version of the contract is deployed in production.


dancreee (approved):

piobab:
I am closing this PR. It is incomplete -> incentives state changed and it requires more state migration. It will be handled in separate PR. Fix for whitelist extracted here https://github.com/mars-protocol/red-bank/pull/316

PR #317, opened 2023-09-05
MP-3299. Oak no. 8.

piobab:
Oak no. 8. requires only version fix during migration but there is much more migration needed after neutron `incentives` changes.


contracts/incentives/src/migrations/v2_0_0.rs
```diff
@@ -0,0 +1,207 @@
...
+                    user: user.to_string(),
+                    account_id: None,
+                    denom: denom.clone(),
+                },
+            )?;
+
+        // Get asset incentive for a denom. It should be available but just in case we don't unwrap
+        let denom_idx = asset_incentives.get(&denom);
+        if let Some(asset_incentive) = denom_idx {
+            // Since we didn't track unclaimed rewards per collateral denom in v1 we add them
```
piobab, line 172: We can `send` these rewards to user wallets but I am not sure if this is good decision. For now I add unclaimed rewards from previous state to first user collateral denom.

piobab:
Pushed fix for schema.

Problem:
```
mars-incentives
Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
status: exit status: 101
stdout: 
stderr: error: There are multiple `mars-incentives` packages in your project, and the specification `mars-incentives` is ambiguous.
Please re-run this command with `-p <spec>` where `<spec>` is one of the following:
  mars-incentives@1.0.0
  mars-incentives@2.0.0
```

Even with specifying versions. Clean & build. Regenerating Cargo.lock etc - nothing helped.
```
cargo run --package mars-incentives:2.0.0 --example schema
error: package(s) `mars-incentives:2.0.0` not found in workspace `MarsProtocol/outposts`
```

contracts/incentives/src/migrations/v2_0_0.rs
```diff
@@ -0,0 +1,286 @@
...
+        .range(deps.storage, None, None, Order::Ascending)
+        .collect::<StdResult<Vec<_>>>()?;
+    v1_state::USER_ASSET_INDICES.clear(deps.storage);
+
+    let mut user_unclaimed_rewards = v1_state::USER_UNCLAIMED_REWARDS
+        .range(deps.storage, None, None, Order::Ascending)
+        .collect::<StdResult<HashMap<_, _>>>()?;
+    v1_state::USER_UNCLAIMED_REWARDS.clear(deps.storage);
+
+    for ((user, denom), user_asset_index) in user_asset_indices {
```
thec00n, line 238: Are we hitting block gas limit here if there are too many users?

piobab: Could be. We will see once we try to migrate the contracts on devnet

dancreee (approved):

dancreee:
lgtm, pretty complex migration :) lets see about gas as theC00n brings up

PR #319, opened 2023-09-07
Revert deposit on behalf of

piobab:
Calc finance uses deposit `on_behalf_of`.

- First commit reverts https://github.com/mars-protocol/red-bank/pull/230 (excluding `yarn` changes). Issue https://delphilabs.atlassian.net/browse/MP-2840
- Second one blocks credit manager contract from using `on_behalf_of`.

### DIFF
--- a/contracts/red-bank/src/deposit.rs
+++ b/contracts/red-bank/src/deposit.rs
@@ -22,16 +22,6 @@ pub fn deposit(
     deposit_amount: Uint128,
     account_id: Option<String>,
 ) -> Result<Response, ContractError> {
-    let user_addr: Addr;
-    let user = if let Some(address) = on_behalf_of.as_ref() {
-        user_addr = deps.api.addr_validate(address)?;
-        User(&user_addr)
-    } else {
-        User(&info.sender)
-    };
-
-    let mut market = MARKETS.load(deps.storage, &denom)?;
-
     let config = CONFIG.load(deps.storage)?;
 
     let addresses = address_provider::helpers::query_contract_addrs(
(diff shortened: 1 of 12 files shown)


contracts/red-bank/src/deposit.rs
```diff
@@ -29,11 +41,18 @@ pub fn deposit(
...
             MarsAddressType::Params,
+            MarsAddressType::CreditManager,
         ],
     )?;
     let rewards_collector_addr = &addresses[&MarsAddressType::RewardsCollector];
     let incentives_addr = &addresses[&MarsAddressType::Incentives];
     let params_addr = &addresses[&MarsAddressType::Params];
+    let credit_manager_addr = &addresses[&MarsAddressType::CreditManager];
+
+    // credit manager can't deposit on behalf of users
```
dancreee, line 52: adding a "why" here might be a good idea for future contributors

dancreee (approved):

contracts/red-bank/src/deposit.rs
```diff
@@ -29,11 +41,18 @@ pub fn deposit(
...
+    if on_behalf_of.is_some() && info.sender == credit_manager_addr {
```
thec00n, line 53: This seems not correct. We want to prevent users from depositing on behalf of the credit-manager, so should be `on_behalf_of == credit_manager_addr`.

contracts/red-bank/src/deposit.rs
piobab: Added original issue description.

contracts/red-bank/src/deposit.rs
piobab: Yeah, you are right! I revised the original issue. Fixed.

thec00n (approved):

PR #321, opened 2023-09-08
MP-3345. MP-3348. Don't allow red-bank users to create alternative account ids.

piobab:
Fixes to @thec00n findings:

- no. 1 (red-bank users with alternative account ids might not be able to withdraw their funds). 
Do not allow red-bank users to create alternative account ids only allow the credit manager to create them by only allowing the credit manager contract to interact with the deposit and withdraw endpoint.

- no. 4 (account_id is not validated in red-bank)
red-bank should not allow users to create account-ids directly and any action should use the default account “”. Only the credit manager should be allowed to create custom account ids because validation is performed for all actions.

### DIFF
--- a/contracts/red-bank/src/deposit.rs
+++ b/contracts/red-bank/src/deposit.rs
@@ -39,6 +39,13 @@ pub fn deposit(
     let params_addr = &addresses[&MarsAddressType::Params];
     let credit_manager_addr = &addresses[&MarsAddressType::CreditManager];
 
+    // Don't allow red-bank users to create alternative account ids.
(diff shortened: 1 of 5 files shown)


piobab:
PR to [revert-deposit-on-behalf-of](https://github.com/mars-protocol/red-bank/tree/revert-deposit-on-behalf-of). Once that branch is merged I will change destination to `release/mars-v2`.

thec00n (approved):

piobab:
Changed destination branch. Pushed rebased commit with latest changes from release/mars-v2

dancreee (approved):

PR #322, opened 2023-09-08
MP-3346. MDR cannot be less than zero

piobab:
The case where max_debt_repayable_denominator underflows need to be handled.

@thec00n I added `mdr_negative` with numbers to spreadsheet.


piobab:
Removed HF check after liquidation.

Aligned with the requirement:
` in extreme cases it can happen that the HF worsens after a liquidation... But those liquidations are still necessary to close those bankrupt positions`

packages/liquidation/src/liquidation.rs
```diff
@@ -54,28 +54,38 @@ pub fn calculate_liquidation_amounts(
         collateral_params,
     )?;
 
+    // All debt is liquidatable: When MDR < 0, it means even repaying the whole debt is not going to be enough
+    // to bring the account back to the THF, so the liquidator should be able to repay all the available debt.
+    // Given the numerator in the MDR formula is always > 0, MDR < 0 happens when the denominator is < 0
+    // (we include the case where it’s 0 given it would make MDR = infinite).
     let max_debt_repayable_numerator = (target_health_factor * health.total_debt_value)
```
dancreee, line 61: this can be shifted to be inside the if statement to save some gas?

dancreee:
Lots of logic in this one good to get a second pair of eyes on it if we can

thec00n (approved):

piobab:
Pushed `move variable to if block`

PR #329, opened 2023-09-26
Bump rust deps. Update osmosis-std

piobab:
- bump rust deps
- use latest osmosis-std 0.19.2 (has a fix for stableswap serde problem). I couldn't prepare proper test case because there is an issue with test-tube and cw-it dependency which breaks some tests. For now prove for correct LSD setup with new osmosis-std in separate branch (just see `query_lsd_price` test case result) https://github.com/mars-protocol/red-bank/tree/fix/stableswap-pool-serde

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.20.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -1996,7 +1997,7 @@ name = "mars-osmosis"
 version = "2.0.0"
 dependencies = [
  "cosmwasm-std",
- "osmosis-std",
+ "osmosis-std 0.19.2",
```
piobab, line 2000: and also fix included here

larry0x (approved):

dancreee (approved):

Issue #336, opened 2023-10-07
Create `mars-types` and `mars-adapters` crates


larry0x:
Currently, we have type definitions (such as `ExecuteMsg`s and `QueryMsg`s) scattered in multiple locations:

- `mars-red-bank-types` crate
- `mars-rover-types` crate
- `mars-health-types` crate
- in `contracts/params/src/types`

These can be coalesced into a single `mars-types` repo as a single source of truth for all type definitions.

Similar with the "adapter" objects, we can create a single `mars-adapters` crate for all of them.

PR #339, opened 2023-10-12
Create `mars-types` crate

larry0x:
Closes: #336

Create a `mars-types` crate as the single source of truth for all type definitions.

Three crates are merged into `mars-types` and deleted:

- `mars-account-nft-types`
- `mars-rover-health-types`
- `mars-rover`

There are 3 instances of duplicate codes which have been fixed (leaving only one copy):

- `mars_rover::adapters::swap` and `mars_red_bank_types::swapper`
- `mars_rover::reentrancy_guard` and `mars_utils::guard`
- `mars_rover::msg` and `mars_red_bank_types::rewards_collector::credit_manager`

The table below compares the previous locations of contents & their locations now:

| Previous                       | Now                                         |
| ------------------------------ | ------------------------------------------- |
| `mars_account_nft_types`       | `mars_types::account_nft`                   |
| `mars_params::types`           | `mars_types::params` (#338)                 |
| `mars_red_bank_types`          | `mars_types`                                |
| `mars_rover_health_types`      | `mars_types::health`                        |
| `mars_rover::adapters`         | `mars_types::adapters`                      |
| `mars_rover::coins`            | deleted (use `cosmwasm_std::Coins` instead) |
| `mars_rover::error`            | `mars_credit_manager::error`                |
| `mars_rover::msg`              | `mars_types::credit_manager`                |
| `mars_rover::reentrancy_guard` | deleted (use `mars_utils::Guard` instead)   |
| `mars_rover::traits`           | `mars_types::traits`                        |
| `mars_zapper_base::msg`        | `mars_types::zapper`                        |


contracts/account-nft/src/execute.rs
```diff
@@ -6,10 +6,12 @@ use cw721_base::{
...
     OwnershipError::{NoOwner, NotOwner},
 };
-use mars_account_nft_types::nft_config::NftConfigUpdates;
-use mars_red_bank_types::oracle::ActionKind;
-use mars_rover::msg::QueryMsg;
-use mars_rover_health_types::{HealthValuesResponse, QueryMsg::HealthValues};
+use mars_types::{
+    account_nft::NftConfigUpdates,
+    credit_manager::QueryMsg,
+    health::{AccountKind, HealthValuesResponse, QueryMsg::HealthValues},
```
larry0x, line 12: I don't like how this code imports a single enum variant (importing `QueryMsg::HealthValues` instead of just `QueryMsg`). It's confusing because I think there is another type also named `HealthValues` (or at least a similar name) so it's unclear which one this is. There are numerous such instances in our codebase. Maybe I will fix it later

contracts/credit-manager/src/deposit.rs
```diff
@@ -27,7 +24,7 @@ pub fn deposit(
 
     assert_sent_fund(coin, received_coins)?;
 
-    received_coins.deduct(coin)?;
+    received_coins.sub(coin.clone())?;
```
larry0x, line 27: We created the `Coins` type for use in Rover and later upstreamed the code to cosmwasm-std. However, some function names are different, for example this one. It also takes a `Coin` instead of `&Coin` so we need to clone it

contracts/credit-manager/src/deposit.rs
```diff
@@ -38,7 +35,7 @@ pub fn deposit(
 
 /// Assert that fund of exactly the same type and amount was sent along with a message
 fn assert_sent_fund(expected: &Coin, received_coins: &Coins) -> ContractResult<()> {
-    let received = received_coins.amount(&expected.denom).unwrap_or_else(Uint128::zero);
+    let received = received_coins.amount_of(&expected.denom);
```
larry0x, line 38: `amount_of` defaults to zero if the denom is not found

contracts/credit-manager/src/error.rs
```diff
@@ -180,4 +174,16 @@ pub enum ContractError {
...
+    Coins(#[from] CoinsError),
+
+    #[error(transparent)]
+    Guard(#[from] GuardError),
+
+    #[error(transparent)]
+    Vault(#[from] VaultError),
+
+    #[error(transparent)]
+    Oracle(#[from] OracleError),
```
larry0x, line 188: I moved `mars_rover::error::ContractError` into the credit manager contract (in `mars_credit_manager::error`) instead of into `mars-types` to stay consistent with other contracts. In order to do this I also had to create two more error types `{Vault,Oracle}Error` which are errors that may happen when interacting with a vault or oracle adapter.

contracts/credit-manager/src/update_config.rs
```diff
@@ -1,14 +1,15 @@
 use cosmwasm_std::{to_binary, CosmosMsg, DepsMut, Env, MessageInfo, Response, WasmMsg};
 use cw721_base::Action;
-use mars_account_nft_types::{msg::ExecuteMsg as NftExecuteMsg, nft_config::NftConfigUpdates};
 use mars_owner::OwnerUpdate;
-use mars_rover::{
-    adapters::rewards_collector::RewardsCollector, error::ContractResult,
-    msg::instantiate::ConfigUpdates,
+use mars_types::{
+    account_nft::{ExecuteMsg as NftExecuteMsg, NftConfigUpdates},
```
larry0x, line 5: Here is another coding style I don't like: renaming imports. Instead of renaming this to `NftExecuteMsg` I would just use `account_nft::ExecuteMsg`. But this is out of scope for this PR

packages/types/src/credit_manager/mod.rs
```diff
@@ -0,0 +1,11 @@
+mod execute;
+mod instantiate;
+mod migrate;
+mod query;
+mod reply;
```
larry0x, line 5: In the original `mars-rover` crate, these are

```rust
pub mod execute;
pub mod instantiate;
pub mod migrate;
pub mod query;
pub mod reply;
```

I removed the `pub mod`, leaving just `mod`. This is because otherwise there is ambiguity in how you import a type:

```rust
// you can either import like this
use mars_types::credit_manager::execute::ExecuteMsg;
// or like this
use mars_types::credit_manager::ExecuteMsg;
```

Typically we want there to be only one way to do something.

packages/types/src/health/mod.rs
```diff
@@ -1,5 +1,6 @@
 mod account;
 mod error;
+#[allow(clippy::module_inception)]
 mod health;
```
larry0x, line 4: This clippy rule basically says we can't have a module named `health` inside a module also named `health`. I think this is a non-issue so ignoring it

contracts/credit-manager/src/error.rs
piobab: I like `transparent` usage

larry0x: `transparent` and `{0}` means exactly the same but the docs recommends `transparent`

contracts/credit-manager/src/state.rs
```diff
@@ -26,7 +26,7 @@ pub const INCENTIVES: Item<Incentives> = Item::new("incentives");
 // Config
 pub const OWNER: Owner = Owner::new("owner");
 pub const MAX_UNLOCKING_POSITIONS: Item<Uint128> = Item::new("max_unlocking_positions");
-pub const REENTRANCY_GUARD: ReentrancyGuard = ReentrancyGuard::new("reentrancy_guard");
+pub const REENTRANCY_GUARD: Guard = Guard::new("reentrancy_guard");
```
piobab, line 29: `ReentrancyGuard` had some logs in Response but I think it is not big problem.

larry0x: Yes it emits a single event attribute. I added it in the credit manager contract.

piobab (approved):
Great job! LGTM

PR #348, opened 2023-11-07
Repay debt with available balance

piobab:
Bound to available balance for `Repay` with `AccountBalance`.


contracts/credit-manager/src/repay.rs
```diff
@@ -9,15 +9,17 @@ use mars_types::credit_manager::{ActionCoin, CallbackMsg::Repay, ExecuteMsg};
...
 };
 
 pub fn repay(deps: DepsMut, account_id: &str, coin: &ActionCoin) -> ContractResult<Response> {
     // Ensure repayment does not exceed max debt on account
     let (debt_amount, debt_shares) =
         current_debt_for_denom(deps.as_ref(), account_id, &coin.denom)?;
-    let amount_to_repay = min(debt_amount, coin.amount.value().unwrap_or(Uint128::MAX));
+    let coin_balance =
+        COIN_BALANCES.may_load(deps.storage, (account_id, &coin.denom))?.unwrap_or_default();
+    let amount_to_repay = min(debt_amount, coin.amount.value().unwrap_or(coin_balance));
```
piobab, line 22: This will bound for `AccountBalance` to available balance. 
If we use `min(debt_amount, current_account_balance(denom), coin.amount.value().unwrap_or(Uint128::MAX))` everything will be bound to available balance.
I can change it but I feel it will be inconsistent with the rest of the usage of `AccountBalance`.

markonmars (approved):

linkielink (approved):

contracts/credit-manager/src/repay.rs
linkielink: But it's still possible to pass exact amounts to `repay`, right?

piobab: Correct. You can still pass Exact or AccountBalance.

`coin.amount.value().unwrap_or(coin_balance)` -> if `Exact` then `coin.amount` else `AccountBalance` then account balance.

piobab:
Sorry guys, there is `RepayForRecipient`. Added the same logic here too (see last commit).

piobab:
Updated migration. Already used for devnet.

PR #352, opened 2024-01-12
MP-3528 pass external routes to swapper

piobab:
Enhancement: External Swapping Routes

This enhancement introduces external routes for swapping with two possible approaches:

1. **Swapper Module Inclusion:**
   - Prepare a swapper module with types and helper functions for inclusion in Credit Manager and Rewards Collector contracts.
   - Contracts depend on a specific package, eliminating the need for a separate Swapper contract.
   - Pros:
     - No additional contract maintenance.
     - Potential reduction in gas consumption.
   - Cons:
     - Increased CM wasm size.
     - Make CM more chain-specific with Astro/Osmo dependency.

2. **Maintain Swapper Contract with Optional Routes:**
   - Keep the Swapper contract as is, allowing optional routes passing.
   - Pros:
     - Smooth migration for Frontend and Bots (Reward Collector).
     - Add fallback for already set routes (if route not provided).
     - Minimal size increase (~20KB) to CM contract.
   - Cons:
     - Swapper maintenance

Decision:
After consideration, the decision is to proceed with option 2:
- Keeps CM chain-agnostic, supporting different routes (Astro or Osmo).
- Minimizes wasm size overhead to keep CM as compact as possible.
- Enables backward compatibility for smooth integration with Frontend and Bots.
- Allows potential switch to option 1 (without swapper contract) if needed after perps completion, potentially reducing gas consumption.

Route struct included in SwapExactIn
Osmo
```
{
    "route": {
        "osmo": {
            "swaps": [
                {
                    "pool_id": 1,
                    "to": "ibc/C4CFF46FD6DE35CA4CF4CE031E643C8FDC9BA4B99AE598E9B0ED98FE3A2319F9"
                },
                {
                    "pool_id": 2,
                    "to": "untrn"
                }
            ]
        }
    }
}
```

Astro
```
{
    "route": {
        "astro": {
            "swaps": [
                {
                    "from": "ibc/3649CE0C8A2C79048D8C6F31FF18FA69C9BC7EB193512E0BD03B733011290445",
                    "to": "ibc/C4CFF46FD6DE35CA4CF4CE031E643C8FDC9BA4B99AE598E9B0ED98FE3A2319F9"
                },
                {
                    "from": "ibc/C4CFF46FD6DE35CA4CF4CE031E643C8FDC9BA4B99AE598E9B0ED98FE3A2319F9",
                    "to": "untrn"
                },
                {
                    "from": "untrn",
                    "to": "ibc/9598CDEB7C6DB7FC21E746C8E0250B30CD5154F39CA111A9D4948A4362F638BD"
                }
            ]
        }
    }
}
```

Astro swapper should be preconfigured (config updated) with router, factory and oracle - more secure to keep them in SC.

Wasm Size Changes reference:
- CM before modification: 889 KB
- CM in the current PR: 913 KB
- R&D scenarios:
  - CM with osmosis feature: 1.1 MB
  - CM with astroport feature: 936 KB
  - CM with both features: 1.2 MB


contracts/swapper/osmosis/tests/tests/helpers/mod.rs
```diff
@@ -138,6 +138,7 @@ pub fn assert_err(actual: RunnerError, expected: impl Display) {
         RunnerError::QueryError {
             msg,
         } => {
+            println!("{}", msg);
```
dancreee, line 141: should this change be in here?

contracts/swapper/osmosis/tests/tests/helpers/mod.rs
```diff
@@ -57,7 +57,7 @@ pub fn swap_to_create_twap_records(
     coin_in: Coin,
     denom_out: &str,
 ) {
-    swap_n_times(app, signer, pool_id, coin_in, denom_out, 120u64);
+    swap_n_times(app, signer, pool_id, coin_in, denom_out, 122u64);
```
dancreee, line 60: also this one

contracts/swapper/osmosis/tests/tests/test_estimate.rs
```diff
@@ -20,11 +19,17 @@ fn error_on_route_not_found() {
...
+                }],
+            })),
         },
     );
-    let err = res.unwrap_err();
-
-    assert_err(err, "No route found from jake to mars");
+    // If no proper route we get ugly error:
+    // "Generic error: Querier contract error: codespace: undefined, code: 1: query wasm contract failed"
+    res.unwrap_err();
```
dancreee, line 32: Seems like we have changed the purpose of this test case, it use to be that no route was found and so would return "No route found from jake to mars"

now it seems we are instead passing an invalid route, shouldn't we keep both test cases? one where there is no route, another where there is an invalid route?

contracts/swapper/osmosis/tests/tests/test_estimate.rs
```diff
@@ -72,6 +62,12 @@ fn estimate_swap_one_step() {
             &QueryMsg::EstimateExactInSwap {
                 coin_in: coin(coin_in_amount.u128(), "uosmo"),
                 denom_out: "uatom".to_string(),
+                route: Some(SwapperRoute::Osmo(OsmoRoute {
+                    swaps: vec![OsmoSwap {
+                        pool_id: pool_atom_osmo,
+                        to: "uatom".to_string(),
+                    }],
+                })),
```
dancreee, line 70: We probably want a number of tests here? 
1. use the default route (the old way we did it by setting route)
2. use the dynamic route (new way of passing the route)
3. set the default route, use the dynamic route and asset dynamic route used?

contracts/swapper/osmosis/tests/tests/test_estimate.rs
```diff
@@ -100,7 +96,7 @@ fn estimate_swap_multi_step() {
         .unwrap()
         .data
         .pool_id;
-    let pool_osmo_mars = gamm
+    let _pool_osmo_mars = gamm
```
dancreee, line 99: can remove if not using?

contracts/swapper/osmosis/tests/tests/test_estimate.rs
```diff
@@ -174,6 +128,18 @@ fn estimate_swap_multi_step() {
...
+                        OsmoSwap {
+                            pool_id: pool_atom_osmo,
+                            to: "uosmo".to_string(),
+                        },
+                        OsmoSwap {
+                            pool_id: pool_osmo_usdc,
+                            to: "uusdc".to_string(),
+                        },
+                    ],
+                })),
```
dancreee, line 142: similar comments here around if we should be having multiple different test cases for default/dynamic/both routes

dancreee (approved):
LGTM, left some minor comments around potential missing test cases etc

markonmars (approved):
Nice! This seems like a clean solution, Addresses the issues I have faced with the bots

Once thought I had while reviewing - Do we check anywhere that the asset we are swapping to (i.e `denom_out`) is the same as is defined in the route? And if they differ does that cause issues?

For instance we set `denom_out`as a whitelisted asset i,e OSMO but the actual route swaps to something different (perhaps not whitelisted), will that tx succeed and is that dangerous?

Apologies in advance if this is already addressed and I missed it.

contracts/credit-manager/src/swap.rs
```diff
@@ -49,7 +53,7 @@ pub fn swap_exact_in(
     let swapper = SWAPPER.load(deps.storage)?;
 
     Ok(Response::new()
-        .add_message(swapper.swap_exact_in_msg(&coin_in_to_trade, denom_out, slippage)?)
+        .add_message(swapper.swap_exact_in_msg(&coin_in_to_trade, denom_out, slippage, route)?)
```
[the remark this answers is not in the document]
piobab: You will get:
```
#[error("No route found from {from} to {to}")]
    NoRoute {
        from: String,
        to: String,
    },
```

If routes is not passed, it tries to read from state, if there is no route then throws NoRoute error

piobab:
> println!("{}", msg);

Very good comment - better ask than we miss something :) 

Yeah, I added validation for off-chain routes to swap and estimate in `base` contract (see screnshots). Validation for routes in states is executed when we set new route. 
Osmosis/Astro swapper contract extends `validate` to check if provided routes are ok (taking denom_in/denom_out and steps):
- https://github.com/mars-protocol/contracts/blob/MP-3528-remove-routes/contracts/swapper/osmosis/src/route.rs#L63
- https://github.com/mars-protocol/contracts/blob/MP-3528-remove-routes/contracts/swapper/astroport/src/route.rs#L140


Zrzut ekranu 2024-01-19 o 11 50 26
Zrzut ekranu 2024-01-19 o 11 49 54

contracts/swapper/osmosis/tests/tests/helpers/mod.rs
piobab: Removed

contracts/swapper/osmosis/tests/tests/helpers/mod.rs
piobab: Added comment. I had to increase number of swaps (one swap move block in osmosis-test-tube).

contracts/swapper/osmosis/tests/tests/test_estimate.rs
piobab: Added such tests to estimate and swap.

contracts/swapper/osmosis/tests/tests/test_estimate.rs
piobab: nah, it is ok :)

contracts/swapper/osmosis/tests/tests/test_estimate.rs
piobab: Added tests for single swap route. It uses the same logic as multi-swaps.

dancreee (approved):
Nice lgtm

PR #354, opened 2024-01-19
Fix inflated collateral

piobab:


markonmars (approved):

contracts/red-bank/src/interest_rates.rs
```diff
@@ -111,12 +111,17 @@ pub fn update_interest_rates(
...
+    let mut current_utilization_rate = if !total_collateral.is_zero() {
         Decimal::from_ratio(total_debt, total_collateral)
     } else {
         Decimal::zero()
     };
 
+    // Limit utilization_rate to 100%.
+    // With the current code it should hopefully never happen that it gets calculated to more than 100%,
+    // but better be safe than sorry.
+    current_utilization_rate = current_utilization_rate.min(Decimal::one());
```
dancreee, line 123: i wonder if it would be better to assert <= 100 here and fail the tx - rather than just capping it at 100% and having a problem maybe go ignored? haven't really thought it through but it would be more defensive

dancreee (approved):

contracts/red-bank/src/interest_rates.rs
piobab: Makes sense! Pushed update.

piobab: Reverted. Calculated amount of debt with interest accured could end up with the situation when utilization is > 100% (not much but it can be).

dancreee (approved):

piobab:
@dancreee @markonmars resolved conflicts and updated CM migration (2.0.2 is already used by swapper routes migration on master)

markonmars (approved):

PR #360, opened 2024-02-21
v2 market(s) query: add underlying collateral and debt

bobthebuidlr:


contracts/red-bank/src/query.rs
```diff
@@ -38,6 +38,27 @@ pub fn query_market(deps: Deps, denom: String) -> StdResult<Option<Market>> {
...
+    let collateral_underlying_amount =
+        get_underlying_liquidity_amount(market.collateral_total_scaled, &market, block_time)?;
+
+    let debt_underlying_amount =
+        get_underlying_debt_amount(market.debt_total_scaled, &market, block_time)?;
+
+    Ok(MarketResponse {
+        debt_underlying_amount,
+        collateral_underlying_amount,
+        collateralization_rate: Decimal::from_ratio(
```
piobab, line 54: Sorry, can we change it to `utilization_rate`? This is what we use in interest rate formula, see:
```
let mut current_utilization_rate = if !total_collateral.is_zero() {
        Decimal::from_ratio(total_debt, total_collateral)
    } else {
        Decimal::zero()
    };
   ```

File: contracts/red-bank/src/query.rs [hunk 2]
```diff
@@ -56,6 +77,43 @@ pub fn query_markets(
         .collect()
 }
 
+pub fn query_markets_v2(
```
piobab, line 80: Could you please implement in similar way as `query_user_collaterals_v2`? Would be good to have new way for pagination. We could avoid spinning new query if there is no new items.

contracts/red-bank/tests/tests/test_query.rs
```diff
@@ -306,3 +311,104 @@ fn query_user_asset_debt() {
         );
     }
 }
+
+#[test]
+pub fn test_query_market_v2() {
```
piobab, line 316: Please remove `test` prefixes, just leave `query_market_v2`

packages/types/src/red_bank/types.rs
```diff
@@ -113,3 +113,13 @@ pub struct UserPositionResponse {
     pub weighted_liquidation_threshold_collateral: Uint128,
     pub health_status: UserHealthStatus,
 }
+
+#[cw_serde]
+pub struct MarketResponse {
```
piobab, line 118: Can we rename it to `MarketV2Response`? It relates for MarketV2 query. It is nice to have query and response naming similar

scripts/types/generated/mars-red-bank/MarsRedBank.types.ts
```diff
@@ -236,7 +247,23 @@ export interface Market {
   liquidity_rate: Decimal
   reserve_factor: Decimal
 }
+export interface MarketResponse {
```
piobab, line 250: Nice that flatten works! :)

packages/types/src/red_bank/types.rs
bobthebuidlr: Done, renamed it

packages/types/src/red_bank/types.rs
```diff
@@ -112,3 +114,15 @@ pub struct UserPositionResponse {
     pub weighted_liquidation_threshold_collateral: Uint128,
     pub health_status: UserHealthStatus,
 }
+
+pub type PaginatedMarketV2Response = PaginationResponse<MarketV2Response>;
+
+#[cw_serde]
+pub struct MarketV2Response {
+    pub collateral_underlying_amount: Uint128,
```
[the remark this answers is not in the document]
bobthebuidlr: Yeah I agree with the renaming to `collateral_total_amount`. I don't feel like it's necessary to create a `MarketV2` struct though

piobab (approved):
LGTM!

PR #362, opened 2024-02-22
add all_vault_utilizations query to CM

bobthebuidlr:


contracts/credit-manager/src/query.rs
```diff
@@ -142,12 +142,50 @@ pub fn query_vault_utilization(
...
+    env: Env,
+    start_after: Option<String>,
+    limit: Option<u32>,
+) -> ContractResult<Vec<VaultUtilizationResponse>> {
+    let params = PARAMS.load(deps.storage)?;
+    let oracle = ORACLE.load(deps.storage)?;
+    let vault_configs = params.query_all_vault_configs(&deps.querier, start_after, limit)?;
+
+    vault_configs
+        .iter()
```
piobab, line 171: If you change order you will avoid addr clone: 
```
Ok(VaultUtilizationResponse {
                utilization: vault_utilization_in_deposit_cap_denom(
                    &deps,
                    &oracle,
                    &vault_config,
                    &env.contract.address,
                )?,
                vault: Vault {
                    address: vault_config.addr,
                }
                .into(),
            })
```

contracts/credit-manager/src/query.rs
```diff
@@ -142,12 +142,50 @@ pub fn query_vault_utilization(
...
+        utilization: vault_utilization_in_deposit_cap_denom(
+            &deps,
+            &oracle,
+            &vault_config,
+            &env.contract.address,
+        )?,
     })
 }
 
+pub fn query_all_vault_utilizations(
```
piobab, line 160: Can we use new pagination?

contracts/credit-manager/src/query.rs
bobthebuidlr: Why is this supposed to make the `clone()` redundant? I just tried and it still not allowing moving of borrowed `addr`.

contracts/credit-manager/src/query.rs
bobthebuidlr: For that to work, I'd have to adjust the `query_all_vault_configs` as well to support the pagination. This means we'd also have to create a v2 version for that one. Should I still do this? Wouldn't hurt I guess, but will just take some more time.

piobab: Up to you. If we create new queries IMO we should go with newest format which will be end goal.

bobthebuidlr: Yeah I do agree here. Let me just add another V12 query for vault_configs to support the new pagination then.

packages/types/tests/test_paginate_query.rs
```diff
@@ -0,0 +1,8 @@
+#[test]
```
bobthebuidlr, line 1: @piobab I'm unsure how to create a mock `Map` that we can use for testing this query

packages/types/src/lib.rs
```diff
@@ -26,3 +30,33 @@ pub struct PaginationResponse<T> {
 pub struct Metadata {
     pub has_more: bool,
 }
+
+pub fn paginate_query<'a, K, T, R, E, F>(
```
bobthebuidlr, line 34: This one only works with `Map`. For sure RB (and probably other contracts) will need to use a `prefix`. Should we just create a second version of this function for prefix variants (feels very duplicate IMO, should be a better way)? @piobab

piobab: Could be tricky to create correct generic types to handle prefix and map in the same function. For now let's do second function as it is in cw-paginate.

packages/types/tests/test_paginate_query.rs
piobab: You can do something like this:
```
#[test]
fn sample_test_case() {
    let mut deps = mock_dependencies();
    let env = mock_env();

    let coin_balances: Map<&str, Uint128> = Map::new("coin_balance");

    coin_balances.save(&mut deps.storage, "key_1", &Uint128::new(100)).unwrap();
    coin_balances.save(&mut deps.storage, "key_2", &Uint128::new(101)).unwrap();
    coin_balances.save(&mut deps.storage, "key_3", &Uint128::new(102)).unwrap();

    let res = paginate_query(&coin_balances, &deps.storage, None, 2, |key, amount| {
        let mapped = CoinBalanceResponseItem {
            account_id: "accoun_1".to_string(),
            denom: key,
            amount,
        };

        Ok::<CoinBalanceResponseItem, StdError>(mapped)
    });
    let res = res.unwrap();
    assert_eq!(res.metadata.has_more, true);
    assert_eq!(res.data.len(), 2);
}
```

contracts/credit-manager/tests/tests/test_utilizations_all_query.rs
```diff
@@ -0,0 +1,256 @@
...
+#[test]
+fn utilizations_are_zero() {
+    let vault_info_1 = unlocked_vault_info();
+    let vault_info_2 = locked_vault_info();
+
+    let mock = MockEnv::new().vault_configs(&[vault_info_1.clone(), vault_info_2]).build().unwrap();
+
+    let res = mock.query_all_vault_utilizations(None, None).unwrap();
+
+    let _ = res.data.iter().map(|vault_utilization| {
```
piobab, line 40: `res.data.iter()` could be used even if in the response are only one item. It doesn't check if there are 2 items.

I would prefer to match like this:
```
assert_eq!(
        res.data,
        vec![
            VaultUtilizationResponse {
                vault: mock.get_vault(&vault_info_1),
                utilization: Coin {
                    denom: vault_info_1.deposit_cap.denom.clone(),
                    amount: Uint128::zero(),
                },
            },
            VaultUtilizationResponse {
                vault: mock.get_vault(&vault_info_2),
                utilization: Coin {
                    denom: vault_info_2.deposit_cap.denom.clone(),
                    amount: Uint128::zero(),
                },
            }
        ]
    );
```

contracts/params/tests/tests/test_query_all_vault_configs_v2.rs
```diff
@@ -0,0 +1,58 @@
...
+        &mock.query_owner(),
+        VaultConfigUpdate::AddOrUpdate {
+            config: vault_config_2,
+        },
+    )
+    .unwrap();
+
+    let vault_configs = mock.query_all_vault_configs_v2(None, Some(1));
+
+    assert_eq!(vault_configs.data.get(0).unwrap().addr, deps.api.addr_validate("vault_1").unwrap())
```
piobab, line 31: You can use:
`Addr::unchecked("vault_1")`

piobab (approved):
LGTM. Only minor changes

contracts/params/tests/tests/test_query_all_vault_configs_v2.rs
piobab, line 31: I will check length also

contracts/params/tests/tests/test_query_all_vault_configs_v2.rs
```diff
@@ -0,0 +1,58 @@
...
+    .unwrap();
+    mock.update_vault_config(
+        &mock.query_owner(),
+        VaultConfigUpdate::AddOrUpdate {
+            config: vault_config_2,
+        },
+    )
+    .unwrap();
+
+    let vault_configs = mock.query_all_vault_configs_v2(None, Some(1));
```
piobab, line 55: Would be nice to check that `Bound::exclusive` works

bobthebuidlr: Done. Also added an extra test case for that.

contracts/credit-manager/tests/tests/test_utilizations_all_query.rs
bobthebuidlr: Agree, adjusted accordingly

PR #366, opened 2024-02-27
Mp 2261 rr neutron

piobab:
Add LSD price source to Wasm (neutron) version of Oracle contract. It uses similar logic to Osmosis but for TWAP we use Astroport.

I use Stride ICA oracle contract as a reference for integration tests. The rest of LSD providers should have the same API.

TODO: migration for older version of Oracle contract on Neutron. This will come after this PR is merged.


contracts/oracle/wasm/src/astroport_twap.rs
```diff
@@ -36,6 +36,11 @@ impl ExecuteTwapSnapshots for WasmOracle<'_> {
                     window_size,
                     tolerance,
                 } => (pair_address, window_size, tolerance),
+                WasmPriceSourceChecked::Lsd {
```
piobab, line 39: @markonmars probably you use Twap price source to know what prices should snapshot. This is new one which has Twap wrapped inside. Please take a look.

dancreee (approved):

PR #367, opened 2024-02-28
Support osmo cosmwasm pool

piobab:
- Add support for CosmWasm Pool (transmuter)
- Use 1:1 pricing (https://github.com/osmosis-labs/transmuter) for CosmWasm pool when estimating out amount. For the rest of the pools we use arithmetic TWAP.

Before/After migration:
Zrzut ekranu 2024-02-28 o 22 18 38

### DIFF
--- a/contracts/swapper/osmosis/src/route.rs
+++ b/contracts/swapper/osmosis/src/route.rs
@@ -2,7 +2,7 @@ use std::fmt;
 
 use cosmwasm_schema::cw_serde;
 use cosmwasm_std::{BlockInfo, CosmosMsg, Decimal, Empty, Env, Fraction, QuerierWrapper, Uint128};
-use mars_osmosis::helpers::{query_arithmetic_twap_price, query_pool, CommonPoolData};
+use mars_osmosis::helpers::{query_arithmetic_twap_price, query_pool, CommonPoolData, Pool};
 use mars_swapper_base::{ContractError, ContractResult, Route};
 use mars_types::swapper::{EstimateExactInSwapResponse, SwapperRoute};
 use osmosis_std::types::osmosis::gamm::v1beta1::MsgSwapExactAmountIn;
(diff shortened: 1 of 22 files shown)


dancreee (approved):

piobab:
@dancreee thx for review. Pushed update for 1:1 pricing when CosmWasm Pool.

dancreee (approved):
lgtm

contracts/swapper/osmosis/src/route.rs
```diff
@@ -201,13 +201,20 @@ fn query_out_amount(
...
-            step.pool_id,
-            &denom_in,
-            &step.token_out_denom,
-            start_time,
-        )?;
+        let pool = query_pool(querier, step.pool_id)?;
+        let step_price = if let Pool::CosmWasm(..) = pool {
+            // TWAP not supported.
+            // This is transmuter (https://github.com/osmosis-labs/transmuter) pool with 1:1 conversion of one asset to another.
+            Decimal::one()
```
markonmars, line 208: Is there any situation where the transmuter will have a successful swap but not return 1? For instance, in the case that it reaches its maximum weight limit for an asset?

Guess this is only a query, just thinking out loud here :smile:

markonmars (approved):

contracts/swapper/osmosis/src/route.rs
piobab: Unfortunatelly not whitelisted :/ https://github.com/osmosis-labs/osmosis/blob/ef49d875fa6cc276d9844840e9feeed36fbdc631/wasmbinding/stargate_whitelist.go#L88

piobab: This query is used to calculate min output amount (slippage).

I updated to use `calc_out_amt_given_in` (see last commit). It will use the same logic as it is for swapping from CosmWasm pool contract (including normalization factors https://github.com/osmosis-labs/transmuter/blob/main/contracts/transmuter/src/swap.rs#L437).

Previous version was also correct but what @markonmars pointed is already handled by this query in the transmuter contract:
```
// ensure there is enough token_out_denom in the pool
        ensure!(
            token_out_pool_asset.amount() >= token_out_amount,
            ContractError::InsufficientPoolAsset {
                required: token_out,
                available: token_out_pool_asset.to_coin()
            }
        );
```

piobab: Verified on newmetric instance:
Zrzut ekranu 2024-03-11 o 19 25 46

piobab: cc @dancreee @markonmars

packages/chains/osmosis/src/helpers.rs
```diff
@@ -186,12 +285,55 @@ pub fn recovered_since_downtime_of_length(
...
+    let res: CalcOutAmtGivenInResponse = querier.query(&QueryRequest::Wasm(WasmQuery::Smart {
+        contract_addr: contract_addr.to_string(),
+        msg: to_json_binary(&CalcOutAmtGivenInRequest {
+            calc_out_amt_given_in: Some(CalcOutAmtGivenIn {
+                token_in: Some(Coin {
+                    denom: token_in.denom.to_string(),
+                    amount: token_in.amount.to_string(),
+                }),
+                token_out_denom: token_out_denom.to_string(),
+                swap_fee: "0".to_string(), // 0 is required by the contract
```
markonmars, line 313: good comment :)

markonmars (approved):
nice :+1:

dancreee (approved):

PR #374, opened 2024-04-22
MP-2342. Pricing xyk lp

piobab:


contracts/oracle/wasm/src/price_source.rs
```diff
@@ -643,6 +671,36 @@ fn query_lsd_price(
     min_price.checked_mul(transitive_price).map_err(Into::into)
 }
 
+/// The calculation of the value of liquidity token, see: https://blog.alphafinance.io/fair-lp-token-pricing/.
+/// This formulation avoids a potential sandwich attack that distorts asset prices by a flashloan.
+///
```
markonmars, line 676: Is this comment still required? Seems like it has been duplicated

markonmars (approved):
LGTM.

Are there any issues with using different price sources? I.e price_source of asset a is a twap while asset b price source is pyth. I imagine this is fine we just end up with some sort of half twap price, but interested in if you had any thoughts on this ?

contracts/oracle/wasm/tests/tests/test_price_source.rs
```diff
@@ -909,3 +922,78 @@ pub fn validate_and_query_lsd_price_source(
...
+        coin0_value = Uint256::from_uint128(coin0.amount) * Decimal256::from(price);
+    }
+    if let Some(price) = other_asset_price {
+        other_assets_price_sources.push((other_denom, fixed_source(price)));
+
+        coin1_value = Uint256::from_uint128(coin1.amount) * Decimal256::from(price);
+    }
+
+    // Calculate expected price
+    let pool_value_u256 = Uint256::from(2u8) * (coin0_value * coin1_value).isqrt();
```
bobthebuidlr, line 984: Why aren't you calling the function on the contract to calculate the xyk price?

bobthebuidlr (approved):

contracts/oracle/wasm/tests/tests/test_price_source.rs
piobab: Hmm take a look on the functions params (I don't have Map to access etc). Moreover I would prefer to don't use the same function in test case as I used in contract code.. If there is an issue I could sneak something to the test case.

contracts/oracle/wasm/src/price_source.rs
piobab: Removed

PR #375, opened 2024-04-23
add `AccountKind` to `Positions` query response

bobthebuidlr:


contracts/credit-manager/tests/tests/test_positions_query.rs
```diff
@@ -0,0 +1,19 @@
+use cosmwasm_std::Addr;
```
piobab, line 1: I think now that we could consider this test case in `test_create_credit_account.rs`

contracts/credit-manager/src/query.rs
```diff
@@ -65,6 +65,7 @@ pub fn query_config(deps: Deps) -> ContractResult<ConfigResponse> {
 pub fn query_positions(deps: Deps, account_id: &str) -> ContractResult<Positions> {
     Ok(Positions {
         account_id: account_id.to_string(),
+        kind: ACCOUNT_KINDS.load(deps.storage, account_id)?,
```
piobab, line 68: I will be more explicit and name it `account_kind`

piobab (approved):

PR #376, opened 2024-04-23
MP-2341. Astro zapper

piobab:
Zap/unzap for Astro.

TODO:
There is no official Astro release for Token Factory yet. We depend on https://crates.io/crates/cw-it in our test cases. It doesn't have Astro Token Factory version support. Apollo have been working on this so we will provide test cases in another PR once ready.


contracts/credit-manager/src/zap.rs
```diff
@@ -48,14 +48,15 @@ pub fn provide_liquidity(
...
-        zapper.estimate_provide_liquidity(&deps.querier, lp_token_out, &updated_coins_in)?;
+        zapper.estimate_provide_liquidity(&deps.querier, lp_token_out, &updated_coins_in, None)?;
     let estimated_min_receive_slippage =
         estimated_min_receive.checked_mul_floor(Decimal::one() - slippage)?;
 
     let zap_msg = zapper.provide_liquidity_msg(
         &updated_coins_in,
         lp_token_out,
         estimated_min_receive_slippage,
+        None,
```
piobab, line 59: Zap/unzap requires astro `pair_addr,` should we pass it from Fronted or add it to `params` config? wdyt @bobthebuidlr @markonmars

bobthebuidlr: Are the `lp_token_out` and the `pair_addr` always 1:1 related? If so, it doesn't make sense for FE to pass both

piobab: Yes, they are connected. I would even deduct `lp_token_out` based on `pair_addr` and vice versa. It looks this is structured `factory/[pair_addr]/astroport/share`

piobab: Pushed update

bobthebuidlr (approved):

markonmars (approved):

PR #380, opened 2024-05-09
Remove route validation for swapper.

piobab:
We have an issue with the pool validation (dynamic routing) on the Osmosis chain for the pool https://celatone.osmosis.zone/osmosis-1/pools/1570.
Contract for the pool:
https://celatone.osmosis.zone/osmosis-1/contracts/osmo1vk3kyfdswfxn7gwprn8fmflfskvp585wm2jnku2up0u7l0l7flcqt7nfqu
Instantiate msg for the contract:
```
{
  "asset_infos": [
    {
      "native_token": {
        "denom": "ibc/498A0751C798A0D9A389AA3691123DADA57DAA4FE165D5C75894505B876BA6E4"
      }
    },
    {
      "native_token": {
        "denom": "ibc/831F0B1BBB1D08A2B75311892876D71565478C532967545476DF4C2D7492E48C"
      }
    }
  ],
  "token_code_id": 0,
  "factory_addr": "osmo1246fnsutktuqqzrru673pqwtt64n288004j5fauyuezwr54llw5sl6drp6",
  "init_params": "eyJhbXAiOiIxMCIsImdhbW1hIjoiMC4wMDAxODEyNSIsIm1pZF9mZWUiOiIwLjAwMDUiLCJvdXRfZmVlIjoiMC4wMDIiLCJmZWVfZ2FtbWEiOiIwLjAwMDIzIiwicmVwZWdfcHJvZml0X3RocmVzaG9sZCI6IjAuMDAwMDAwMiIsIm1pbl9wcmljZV9zY2FsZV9kZWx0YSI6IjAuMDAwMTQ2IiwibWFfaGFsZl90aW1lIjo2MDAsInByaWNlX3NjYWxlIjoiMy45MiJ9"
}
```
We don't support such pool in our contracts. Deserialization will fail here:
https://github.com/mars-protocol/contracts/blob/master/packages/chains/osmosis/src/helpers.rs#L128

Validating pools was logical for pre-configured routing to prevent issues during swapping if the route was incorrectly set. However, with dynamic routing on Mars smart contracts, there's no need for route validation. Instead, Osmosis/Astro DEXes will validate the route during the swap execution. Additionally, this is problematic for us because we cannot anticipate what optimal routing will be provided by Osmosis/Astro, necessitating us to ensure support for all types of pools in advance.

### DIFF
--- a/contracts/swapper/base/src/contract.rs
+++ b/contracts/swapper/base/src/contract.rs
@@ -175,10 +175,7 @@ where
 
         // if route is not provided, use the default route from state
         let route = match route {
-            Some(route) => {
-                let route = R::from(route, config)?;
-                route
-            }
+            Some(route) => R::from(route, config)?,
             None => self.get_route(deps, &coin_in.denom, &denom_out)?,
         };
         route.estimate_exact_in_swap(&deps.querier, &env, &coin_in)
@@ -207,9 +204,7 @@ where
             Some(route) => {
                 let config = self.query_config(deps.as_ref())?;
 
-                let route = R::from(route, config)?;
-
-                route
+                R::from(route, config)?
             }
             None => self
                 .routes
--- a/contracts/swapper/base/src/contract.rs
+++ b/contracts/swapper/base/src/contract.rs
@@ -177,7 +177,6 @@ where
         let route = match route {
             Some(route) => {
                 let route = R::from(route, config)?;
-                route.validate(&deps.querier, &coin_in.denom, &denom_out)?;
                 route
             }
             None => self.get_route(deps, &coin_in.denom, &denom_out)?,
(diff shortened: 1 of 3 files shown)


markonmars (approved):

contracts/swapper/base/src/contract.rs
```diff
@@ -175,11 +175,7 @@ where
 
         // if route is not provided, use the default route from state
         let route = match route {
-            Some(route) => {
-                let route = R::from(route, config)?;
-                route.validate(&deps.querier, &coin_in.denom, &denom_out)?;
```
dancreee, line 180: is this the only place using the validate fn on Route? if so should we clean that up too

dancreee (approved):

contracts/swapper/base/src/contract.rs
piobab: I don't understand :D Wdym? It is removed for provided routing but it is left for pre-configured (fallback).

PR #383, opened 2024-05-14
MP-2390. Account custom token

piobab:


contracts/account-nft/src/helpers.rs
```diff
@@ -0,0 +1,23 @@
+use crate::error::ContractError;
+
+/// token_id follows Twitter username/handle rules:
```
piobab, line 3: We can tune it later but for now I inspired from Twitter handle

markonmars (approved):

piobab:
@bobthebuidlr @markonmars pushed one commit to fix issue with overlapping custom token id with automatically generated.

markonmars (approved):

PR #384, opened 2024-05-15
MP-2381. Fund manager acc

piobab:
FundManager account has two actors - wallet which is the owner of the account (can’t deposit and withdraw from the account) and vault which can do whatever he wants on the account.

I changed the managed vault initialization logic. In the hackathon version, the credit manager had to transfer initializations to the contract vault, which created a tight connection between contracts. See here:
Zrzut ekranu 2024-05-15 o 11 04 42
I am in favor of the Credit Manager knowing as little as possible about who uses it. Fund Manager will initialize the vault contract and in Credit Manager we will only link the account to the vault address and grant appropriate permissions for the Wallet and Fund Manager Vault. Basically we need:
- 1 msg to instantiate a vault
- 1 msg to mint a credit account and bind the account with Vault addr

Additionally, the Fund Manager can decide whether the Vault should be immutable or not.

### DIFF
--- a/contracts/credit-manager/src/execute.rs
+++ b/contracts/credit-manager/src/execute.rs
@@ -352,15 +352,15 @@ fn validate_account(
     deps: &DepsMut,
     info: &MessageInfo,
     acc_id: &String,
-    actions: &Vec<Action>,
+    actions: &[Action],
 ) -> Result<(), ContractError> {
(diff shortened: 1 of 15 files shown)


markonmars (approved):

piobab:
@markonmars cherry-picked commits to proper target. Commits are the same as previously:
Zrzut ekranu 2024-05-21 o 10 33 52

PR #391, opened 2024-05-28
MP-2384. Withdraw from vault

piobab:
Withdraws for Manager Vault
* Users can perform an unlock request anytime, this will lock their withdraw to 24hrs (can be configured).
Unlock request contains number of shares (vault tokens) the user wants to withdraw (actual vault tokens should be send with Withdraw request, not Unlock). The user can preview how much USDC will get back (we calculate FundManager account value by summing all collaterals and subtracting Debt - valued in USDC). The real amount of USDC after 24h can be different because Fund Manager can for example underperform.
* After which they will be able to withdraw by burning their shares (vault tokens) and grabbing USDC cash. Withdraw request should contain correct amount of vault tokens specified  previously in Unlock request.
* If the account doesn’t have enough USDC, it will try to unlend and if still not enough borrow against its collateral
* The idea is that the Manager at any point in time has enough knowledge of how the total amount of shares queued to be withdrawn in the next 24hrs and can make sure there is enough USDC available


markonmars (approved):
Nice, code looks good :+1:  

One thing I wonder though is if there is some sort of value leakage due to that fact that we allow vault participants to claim usd based on current realised + unrealised pnl, as there is usually some cost associated with realising pnl. For instance a high levered vault on something like a delta neutral strategy might have a high cost of closing the position, so there could be some gain to exiting the vault before the PnL is realised / position is closed. 

Haven't thought this through too much though, so maybe I'm just splitting hairs :smile:

piobab:
> Nice, code looks good 👍
> 
> One thing I wonder though is if there is some sort of value leakage due to that fact that we allow vault participants to claim usd based on current realised + unrealised pnl, as there is usually some cost associated with realising pnl. For instance a high levered vault on something like a delta neutral strategy might have a high cost of closing the position, so there could be some gain to exiting the vault before the PnL is realised / position is closed.
> 
> Haven't thought this through too much though, so maybe I'm just splitting hairs 😄

I am not sure how we can handle this. We value assets (vault including) the same way for liquidation logic. There is no API to handle the cost of closing..

PR #392, opened 2024-05-29
added stableswap pricing

bobthebuidlr:


contracts/oracle/wasm/src/lp_pricing.rs
```diff
@@ -138,3 +141,83 @@ pub fn compute_pcl_lp_price_real(
...
+    coin1_price: Decimal,
+    coin0_decimals: u8,
+    coin1_decimals: u8,
+    total_shares: Uint128,
+    curve_invariant: Uint128,
+) -> ContractResult<Decimal> {
+    // StableSwap pool lp price calculation:
+    //    virtual_price = curve_invariant / total_shares
+    //    lp_price = virtual_price * min(coin0_price, coin1_price)
+    let virtual_price = Decimal::from_ratio(curve_invariant.checked_div(total_shares)?, 1u128);
```
piobab, line 200: `let virtual_price = Decimal::checked_from_ratio(curve_invariant, total_shares)?;`

contracts/oracle/wasm/src/lp_pricing.rs
```diff
@@ -138,3 +141,83 @@ pub fn compute_pcl_lp_price_real(
...
+
+    // The curve_invariant takes on the precision of the asset with the greatest precision.
+    // https://github.dev/astroport-fi/astroport-core/blob/a0a71af801be3f72c64b81f798e1b0805cf0f594/contracts/pair_stable/src/contract.rs#L91
+    // E.g. a stable pool with asset1_decimals = 6 and asset2_decimals = 18 will have a
+    // curve_invariant with 18 decimals.
+    let greatest_precision = max(coin0_decimals, coin1_decimals);
+
+    let coin0_price_scaled =
```
piobab, line 208: Can we add something like this to illustrate with some numbers:

```
1 BTC = 65 000 USD

We have to provide prices for the whole unit:
10^8 ubtc = 65 * 10^9 uusd
1 ubtc = 65 * 10^9 / 10^8 uusd
1 ubtc = 650 uusd
If we multiply by BTC decimals:
1 BTC = 650 * 10^8 uusd
1 BTC = 65 * 10^9 uusd
```

contracts/oracle/wasm/src/lp_pricing.rs
```diff
@@ -138,3 +141,83 @@ pub fn compute_pcl_lp_price_real(
...
+        Decimal256::from(coin0_price) * Decimal256::from_str("10")?.pow(u32::from(coin0_decimals));
+    let coin1_price_scaled =
+        Decimal256::from(coin1_price) * Decimal256::from_str("10")?.pow(u32::from(coin1_decimals));
+
+    let lp_price_256_scaled =
+        Decimal256::from(virtual_price).checked_mul(min(coin0_price_scaled, coin1_price_scaled))?;
+
+    // The price needs to be adjusted with the greatest_precision to denominate correctly in
+    // uusd per share
+    let lp_price_256 = lp_price_256_scaled
```
piobab, line 218: You can use `as` so it would be `pow(greatest_precision as u32)`

contracts/oracle/wasm/src/lp_pricing.rs
```diff
@@ -138,3 +141,83 @@ pub fn compute_pcl_lp_price_real(
...
+        .checked_div(Decimal256::from_str("10")?.pow(u32::from(greatest_precision)))?;
```
piobab, line 219: IMO this is incorrect.. 
Astro does something like:
```
fn to_uint128_with_precision(&self, precision: impl Into<u32>) -> StdResult<Uint128> {
        let value = self.atomics();
        let precision = precision.into();

        value
            .checked_div(10u128.pow(self.decimal_places() - precision).into())?
            .try_into()
            .map_err(|o: ConversionOverflowError| {
                StdError::generic_err(format!("Error converting {}", o.value))
            })
    }
```
so in order to reverse division by `10u128.pow(self.decimal_places() - precision)` in Astro, we have to multiply on our side by similar formula.

Or do I miss something?

piobab: We discussed this and it looks like only `self.decimal_places()` needs confirmation.

piobab (approved):

contracts/oracle/wasm/src/lp_pricing.rs
bobthebuidlr: Added an explanation in the comments.

PR #395, opened 2024-06-03
Fix astro deps in incentives.

piobab:


packages/testing/src/lib.rs
```diff
@@ -2,8 +2,8 @@
 
 extern crate core;
 
-#[cfg(feature = "astroport")]
 pub mod astroport_incentives_querier;
+#[cfg(feature = "astroport")]
```
piobab, line 6: This was one of the issue. Cfg was moved.

contracts/incentives/tests/tests/test_claim_astro_lp_rewards.rs
```diff
@@ -1,8 +1,8 @@
+use astroport_v5::asset::Asset;
```
piobab, line 1: Updated to use correct astro version for incentives integration.

markonmars (approved):

PR #397, opened 2024-06-04
Astro LP liquidation.

piobab:


contracts/credit-manager/src/execute.rs
```diff
@@ -260,6 +261,12 @@ pub fn dispatch_actions(
                         position_type,
                     },
                 }),
+                LiquidateRequest::AstroLp(lp_denom) => callbacks.push(CallbackMsg::Liquidate {
```
piobab, line 264: Should we name it maybe `StakedLp` instead of `AstroLp`?

contracts/health/src/compute.rs
```diff
@@ -41,6 +42,7 @@ pub fn compute_health(
         .chain(debt_denoms)
         .chain(lend_denoms)
         .chain(vault_base_token_denoms)
+        .chain(staked_lp_denoms)
```
piobab, line 45: @markonmars added what I wrote on TG

contracts/credit-manager/src/execute.rs
markonmars: hmm yea maybe that makes more sense. OR even `stakedAstroLp`?

markonmars (approved):

contracts/credit-manager/src/execute.rs
piobab: Yeah makes more sense. Should we also change in `Positions` response from `staked_lp` to `staked_astro_lp` to be more explicit?
cc @bobthebuidlr

bobthebuidlr: Yes I think we should use `astro` prefix for everything relating to only astro or the `astro_incentives` state. In this case `staked_astro_lp` is the better fit imo.

bobthebuidlr (approved):

contracts/credit-manager/src/execute.rs
piobab: We use plurals for lends, debts. Should we use the same for lp?

piobab: I will rename to `staked_astro_lps` in next commit

piobab: Renamed

piobab:
Cherry-picked commits and changed staked_lp in positions to `staked_astro_lps`

PR #405, opened 2024-06-11
Improve vault unlocks

piobab:
Use `created_at` as part of the `unlocks` key so we can paginate in ASC/DESC order.


piobab:
@bobthebuidlr LMK if such API makes more sense. We can see unlocks in sorted order

bobthebuidlr (approved):

contracts/vault/src/execute.rs
```diff
@@ -195,10 +200,8 @@ pub fn redeem(
     }
 
     // clear state if no more unlocking positions
```
markonmars, line 202: does this comment still make sense?

markonmars (approved):

contracts/vault/src/execute.rs
piobab: Fixed

PR #406, opened 2024-06-14
MP-2435. Keep claimed rewards in the account

piobab:
- Keep claimed rewards in credit account
- Remove deposits check from HLS rule. It is handled in Health Computer collaterals calculation (see diagram). Only correlated deposits to debt can contribute to collateral value.
- Fix missing `staked_astro_lps` for HLS
- It solves issue for sending rewards to Fund Manager wallet (they are just kept in the account)

High level overview (https://excalidraw.com/#json=zzlhNw1NkvfV8axfJDGl9,Ec_67y7m-_3feub_xC4NSA):
hls_rules

### DIFF
--- a/contracts/credit-manager/src/hls.rs
+++ b/contracts/credit-manager/src/hls.rs
@@ -1,5 +1,5 @@
 use cosmwasm_std::{Deps, Response};
-use mars_types::{health::AccountKind, params::HlsAssetType};
+use mars_types::{credit_manager::Positions, health::AccountKind, params::HlsAssetType};
 
 use crate::{
     error::{ContractError, ContractResult},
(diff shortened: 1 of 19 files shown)


packages/health-computer/src/health_computer.rs
```diff
@@ -566,6 +565,55 @@ impl HealthComputer {
...
 
+    fn coin_contribute_to_collateral(&self, coin: &Coin) -> HealthResult<Option<&AssetParams>> {
+        let Some(asset_params) = self.denoms_data.params.get(&coin.denom) else {
+            // If the coin is not found (whitelisted), it is not considered for collateral
+            return Ok(None);
+        };
+
+        match self.kind {
+            AccountKind::HighLeveredStrategy => {
+                // HLS should have 0 or 1 debt denom in the account. If there are more than 1 we can safely calculate the collateral value
```
markonmars, line 576: should this be `can not safely calculate` collateral value?

markonmars (approved):

bobthebuidlr (approved):

contracts/credit-manager/src/hls.rs
```diff
@@ -9,15 +9,24 @@ use crate::{
 
 pub fn assert_hls_rules(deps: Deps, account_id: &str) -> ContractResult<Response> {
     // Rule #1 - There can only be 0 or 1 debt denom in the account
-    let positions = query_positions(deps, account_id)?;
+    let Positions {
+        // destruct Positions so whenever we add new positions we don't forget to add them here
```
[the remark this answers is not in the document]
piobab: @markonmars is it clear or should we update the comment? We missed `staked_astro_lps` here.. with destruction feature the compiler will just inform us in this place.

packages/health-computer/src/health_computer.rs
piobab: I don't understand.. wdym?

markonmars: the comment says ` If there are more than 1 we can safely calculate the collateral value` - but I thought this was a HLS rule where we could not have more than 1 debt?

piobab: Yes, and that's correct. See second part of the comment `because the rule will be checked in the Credit Manager contract and won't allow more than 1 debt denom in the account.`. So even if frontend uses more than one debt for HLS calculations then SC will assert correctness of state withy 1 debt. What comment would be much clear?

markonmars: Yes that makes sense to me. Sorry I missread :)

PR #411, opened 2024-06-20
Fix Oak no. 2 - incorrect LP total state accounting

markonmars:


contracts/incentives/tests/tests/test_claim_astro_lp_rewards.rs
```diff
@@ -434,3 +434,273 @@ fn assert_only_credit_manager() {
...
+            amount: mars_types::credit_manager::ActionAmount::AccountBalance,
+        },
+    )
+    .unwrap();
+
+    // LP in incentives = 50
+    // User a = 50
+    // User b = 0
+    assert_eq!(
+        ASTRO_TOTAL_LP_DEPOSITS.may_load(&deps.storage, lp_denom).unwrap(),
```
piobab, line 569: `load`

contracts/incentives/tests/tests/test_claim_astro_lp_rewards.rs
```diff
@@ -434,3 +434,273 @@ fn assert_only_credit_manager() {
...
+        )
+        .unwrap()
+        .lp_coin
+        .amount,
+        Uint128::new(0u128)
+    );
+}
+
+#[test]
+fn stake_correctly_updates_lp_states() {
```
piobab, line 599: I think you can delete this test case. It does almost the same as above. Maybe just add one extra deposit in above test case and rename the test case from `unstake_correctly_updates_lp_states` to `lp_states_correctly_updated`.

piobab (approved):

bobthebuidlr (approved):

PR #413, opened 2024-06-21
Fix Oak no. 6 - incorrect Unstake exact LP

markonmars:


contracts/credit-manager/src/unstake_astro_lp.rs
```diff
@@ -36,7 +36,7 @@ pub fn unstake_lp(
                     available: lp_position.lp_coin.amount,
                 });
             } else {
-                lp_position.lp_coin.amount.checked_sub(amt)?
+                amt
```
piobab, line 39: Can we change `new_amount` variable to `amount_to_unstake`? It will be much explicit that it is amount to unstake and not new amount as a balance in CM.

markonmars: much better

piobab (approved):

bobthebuidlr (approved):

PR #416, opened 2024-06-24
add staked lp to prop tests

markonmars:


packages/health-computer/tests/tests/helpers/prop_test_strategies.rs
```diff
@@ -273,8 +273,9 @@ pub fn random_health_computer() -> impl Strategy<Value = HealthComputer> {
             random_debts(denoms_data.clone()),
             random_coins(denoms_data.clone()),
             random_vault_positions(vaults_data.clone()),
+            random_coins(denoms_data.clone()),
         )
-            .prop_map(move |(kind, deposits, debts, lends, vaults)| HealthComputer {
+            .prop_map(move |(kind, deposits, debts, lends, vaults, staked_astro_lps)| HealthComputer {
```
piobab, line 278: Maybe we should generate with proper LP name structure for astro?

bobthebuidlr: Agree. I think we shouldn't use the same denoms for lp positions.

markonmars: added denom generation that matches astroport lp share tokens format

piobab:
@markonmars I pushed commit fixing fmt issue

piobab (approved):

bobthebuidlr (approved):

PR #417, opened 2024-06-24
fix issues oak audit no 16

bobthebuidlr:


contracts/vault/tests/tests/test_instantiate.rs
```diff
@@ -146,6 +146,43 @@ fn cannot_instantiate_with_invalid_performance_fee() {
     );
 }
 
+#[test]
+fn cannot_instantiate_with_zero_cooldown_period() {
```
piobab, line 150: Can we add extra test for invalid denom?

piobab: Added

piobab (approved):

markonmars (approved):

PR #418, opened 2024-06-24
Fix Oak no. 9

piobab:


contracts/vault/src/execute.rs
```diff
@@ -101,7 +103,7 @@ pub fn deposit(
     let event = Event::new("deposit").add_attributes(vec![
         attr("action", "mint_vault_tokens"),
         attr("recipient", vault_token_recipient.to_string()),
-        attr("mint_amount", vault_tokens),
+        attr("vault_tokens_minted", vault_tokens),
```
piobab, line 106: I updated naming to attrs to be more consistent.

bobthebuidlr (approved):

markonmars (approved):

PR #423, opened 2024-07-09
MP-2518. Allow to provide min amount for swap

piobab:
We used TWAP and oracle prices to prevent pool manipulation. However, since we can now trade any assets (https://github.com/mars-protocol/contracts/pull/381), we don't have a reliable price source to calculate the `min_amount` in the smart contract (SC).
If we use a simulation request on Astro, the user won't know how much they will receive. By calculating it outside of the smart contract and displaying it to the user, they will know the exact amount they will receive.

**This is braking change**

`Swap` action on credit manager:
```diff
SwapExactIn {
    coin_in: ActionCoin,
    denom_out: String,
-   slippage: Decimal,
+   min_receive: Uint128,
    route: Option<SwapperRoute>,
}
```
`SwapAsset` on rewards-collector:
```diff
SwapAsset {
    denom: String,
    amount: Option<Uint128>,
    safety_fund_route: Option<SwapperRoute>,
    fee_collector_route: Option<SwapperRoute>,
+   safety_fund_min_receive: Option<Uint128>,
+   fee_collector_min_receive: Option<Uint128>,
}
```

### DIFF
--- a/contracts/credit-manager/tests/tests/test_swap.rs
+++ b/contracts/credit-manager/tests/tests/test_swap.rs
@@ -171,6 +171,45 @@ fn user_has_zero_balance_for_swap_req() {
 }
 
 #[test]
+fn min_receive_not_reached() {
+    let osmo_info = uosmo_info();
+    let atom_info = uatom_info();
+
+    let user = Addr::unchecked("user");
+    let mut mock = MockEnv::new()
+        .set_params(&[osmo_info.clone(), atom_info.clone()])
+        .fund_account(AccountToFund {
+            addr: user.clone(),
+            funds: vec![Coin::new(10_000u128, osmo_info.denom.clone())],
+        })
+        .build()
+        .unwrap();
+    let account_id = mock.create_credit_account(&user).unwrap();
+
+    let swap_amt = Uint128::new(10_000);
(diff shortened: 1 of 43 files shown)


markonmars (approved):

linkielink (approved):
LGTM

contracts/credit-manager/tests/tests/test_swap.rs
```diff
@@ -173,44 +171,42 @@ fn user_has_zero_balance_for_swap_req() {
...
+                    swaps: vec![OsmoSwap {
+                        pool_id: 101,
+                        to: atom_info.denom.clone(),
+                    }],
+                })),
+            },
+        ],
+        &[osmo_info.to_coin(swap_amt.u128())],
     )
+    .unwrap_err();
```
bobthebuidlr, line 209: Isn't there a better way we can tewst for the `min_receive` error? I know we're dependent on the dex though

bobthebuidlr (approved):

contracts/credit-manager/tests/tests/test_swap.rs
piobab: I tried but I couldn't read it.

PR #424, opened 2024-07-15
remove `min` from pcl lp price model

bobthebuidlr:


piobab:
This is what we are changing:
Zrzut ekranu 2024-07-9 o 09 42 29

piobab (approved):

markonmars (approved):

PR #428, opened 2024-08-03
MP-2425 Burn fee collector denom rewards

dancreee:
I've removed the code that would distribute via ibc transfer the fee_collector_denom, and replaced it with a burn msg.

This removes the ability for us to do buy & distribute, but this will be a different implementation for DAODAO anyway (currently not supported for tokenFactory tokens AFAIK) so makes sense to remove it and later re-implement it.

We still do the ibc transfer of the safety fund denom to Mars Hubs Safety Fund module on both Osmosis and Neutron outposts. Once we change the Safety Fund to flow to the DAODAO treasury we can probably pull out the Neutron specific impl. of ibc_transfer. 

After completing this work i'm not so sure it's actually worth rolling out until we implement more of the upcoming work to the rewards collector. 

This is because:

1. This buy & burn impl. only works on Neutron/native tokenFactory chain, we couldn't deploy this version on Osmosis it would incorrectly burn the IBC denom rather than the source denom. 
2. There is another refactor in tokenomics phase 2 that should rework the mars address type "SafetyFund" into "Treasury", and sending the safety fund denom to the DAODAO gov contract address. This would also mean we can remove the ibc_transfer stuff for the Neutron implementation as mentioned above.
3. There is an addition in tokenomics phase 2 - We need to add another bucket to protocol revenue called something like "RevenueShare". This is to support our 10% protocol revenue share with Neutron. I'd propose we actually create the ability to protocol revenue share with multiple addresses, such as the Mars Foundation in the future if it's voted by gov or another entity if we do similar on another raise. 

So we could end up with the following protocol revenue buckets that can be configured by governance:

1. Buy & Burn Rate (different impl for Neutron and Osmosis, Neutron easy has the native bank burn msg, more sophisticated ibc transfer to contract + IBC hook to call burn on contract)
2. Treasury Rate (different impl. for Neutron and Osmosis, Neutron has bank send, Osmosis has ibc transfer)
3. Revenue Share Rate (could support two types here, bank send and ibc transfer rev share, but initially just need the bank send)

These 3 should add up to 100%. Additionally within Revenue Share we could have n number of shares with a rate (that add up to 100% of the Revenue Share Rate).

In the future we can also add Buy & LP bucket and Buy & Distribute bucket. Because so many I think the code design needs to be re-worked a bit

TODO

- [x] Contract Migration


packages/types/src/rewards_collector.rs
```diff
@@ -205,7 +205,12 @@ pub enum QueryMsg {
...
 
 #[cw_serde]
-pub enum MigrateMsg {
+pub enum OsmosisMigrateMsg {
     V1_0_0ToV2_0_0 {},
     V2_0_0ToV2_0_1 {},
 }
+
+#[cw_serde]
+pub enum NeutronMigrateMsg {
```
dancreee, line 214: Not sure if this is a clean way to do this, with separate migrate msgs for separate impl.

I had a look elsewhere but didn't see SC where both impl. had migrated

contracts/rewards-collector/neutron/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name          = "mars-rewards-collector-neutron"
-version       = { workspace = true }
+version       = "2.0.2"
```
piobab, line 3: Please leave workspace version.

contracts/rewards-collector/neutron/src/lib.rs
```diff
@@ -92,7 +90,13 @@ pub mod entry {
     }
 
     #[entry_point]
-    pub fn migrate(deps: DepsMut, _env: Env, _msg: Empty) -> ContractResult<Response> {
-        migrations::v2_0_0::migrate(deps)
+    pub fn migrate(
```
piobab, line 93: Migration was already done from 1.2.0 to current workspace version. You can rollback this change,

packages/types/src/rewards_collector.rs
piobab: TBH it doesn't make sense to keep old migrate msgs when we only do simple version/name checks and update to new version. We rename the msg to required one. Moreover it is hard to use old msgs in tests because we loose info about previous version (we change it in Cargo.toml so the contract will use this new version for all tests).

contracts/rewards-collector/neutron/src/migrations/mod.rs
```diff
@@ -1 +1 @@
-pub mod v2_0_0;
+pub mod v2_0_2;
```
piobab, line 1: Right now on ntrn we have 2.1.0. We should update migration

PR #429, opened 2024-08-07
Managed vault - add share price to VaultInfo query

piobab:


contracts/vault/src/msg.rs
```diff
@@ -98,6 +98,15 @@ pub struct VaultInfoResponseExt {
 
     /// Performance fee configuration
     pub performance_fee_config: PerformanceFeeConfig,
+
+    /// Total base tokens in the vault
+    pub total_base_tokens: Uint128,
```
piobab, line 103: @bobthebuidlr @Subham2804 do you think it is good place for these data?

Subham2804: Looks good to me (will require no other query for backend API implementation)

Subham2804 (approved):

contracts/vault/src/msg.rs
Subham2804: looks good to me

piobab: @Subham2804 do you index the rest from this query? I mean title etc

Subham2804: we do use subtitles , vault name etc from this query

bobthebuidlr: Yes this looks like a fine place to me too

contracts/vault/src/query.rs
```diff
@@ -14,16 +14,35 @@ use crate::{
...
+    // It means that the vault is not binded to any account yet.
+    let vault_account_id_opt = VAULT_ACC_ID.may_load(deps.storage)?;
+    let mut total_base_tokens = Uint128::zero();
+    let mut share_price = None;
+    if vault_account_id_opt.is_some() {
+        total_base_tokens = total_base_tokens_in_account(deps)?;
+        share_price = if total_vault_tokens.is_zero() {
+            None
+        } else {
+            Some(Decimal::checked_from_ratio(total_base_tokens, total_vault_tokens)?)
```
bobthebuidlr, line 30: Maybe we need to place a comment that the share_price is always denominated in the base token denom, and not usdc per se.

bobthebuidlr (approved):

contracts/vault/src/query.rs
piobab: @bobthebuidlr please take a look on latest commit.

```
    /// Current share price, representing the value of one vault token in terms of the base token.
    /// It is calculated as the ratio of total base tokens in the vault to the total supply of vault tokens.
    /// Denominated as a decimal value:
    /// `share_price = total_base_tokens / total_vault_tokens`
    /// This share price allows users to determine how many base tokens can be redeemed per vault token:
    /// `base_tokens = vault_share * vault_tokens`
    pub share_price: Option<Decimal>,
```

bobthebuidlr: Beautiful!

PR #434, opened 2024-08-30
MP-2547 burn contract

dancreee:
Contract to be deployed on Neutron for burning the token.

I've added some state/queries so for now we can easily keep track of total burn amounts.

Originally I was going to just use this contract from Osmosis to do the native burn, but I think it makes sense to also route the burns from Neutron through this contract so we have 1 central place that does the accounting for the total burn amounts.

Eventually when a mature indexing solution is in place to account for the burns and show historical data we could simplify the contract removing the state/queries


contracts/burn/src/contract.rs
```diff
@@ -0,0 +1,117 @@
...
+    match msg {
+        ExecuteMsg::BurnFunds {
+            denom,
+        } => execute_burn_funds(deps, env, denom),
+    }
+}
+
+pub fn execute_burn_funds(deps: DepsMut, env: Env, denom: String) -> ContractResult<Response> {
+    let contract_address = env.contract.address;
+    let balance = deps.querier.query_balance(contract_address, denom.clone())?;
```
markonmars, line 46: I wonder if we should check that no other funds were mistakenly sent with the msg? 

For instance if I send [100uosmo, 100uatom] and  pass uosmo as a denom, I will lose the uatom right?  

Might be an over optimisation given that this is a burn address so you should expect to lose funds you send to it, but could prevent accidental mistakes.

markonmars: Another nit: I think in general using `must_pay` is probably best practice here but I don't actually see any issue with the way it is implemented. It might actually be desirable that we just burn the balance of the contract for that denom

markonmars (commented):
LGTM! Left a couple minor comments for consideration, but no strong opinion so feel free to disregard.

contracts/burn/src/contract.rs
dancreee: I don't necessarily want to tie into the same tx a bank send with a wasm burn always, imo keep it flexible so you can do these separately (e.g. if there is an issue with IBC hooks, we can separate out the IBC send with the wasm execute to burn)

We could be more defensive here and set a burn denom and reject any other denom sent to the contract, although imo this contract could be useful for others so I think it could be an over-optimisation in this case, but i'm pretty 50:50 on that so could be swayed

dancreee:
Cancelling this PR and deleting the branch - Neutron won't support this type of burn - opting for the most basic burn possible which is using the Neutron zero address (a simple bank send from Neutron rewards collector, and an already supported IBC-send from Osmosis rewards collector)

PR #436, opened 2024-10-10
Clean incentives state when users unstake

markonmars:
When users unstake from Astroport via Mars incentives contract, we should clean up the state if the user has no more position.

We should also do the same if the denom has no positions (i.e TOTAL_STAKED_LP = 0)


contracts/incentives/src/astro_incentives.rs
```diff
@@ -232,29 +236,75 @@ fn decrement_staked_lp(
...
 
+    let new_amount = existing_amount.checked_sub(lp_coin.amount)?;
+    if new_amount.is_zero() {
+        ASTRO_USER_LP_DEPOSITS.remove(store, key);
+
+        // Get all incentives for (user, lp_token_denom) key
+        let prefix = USER_ASTRO_INCENTIVE_STATES.prefix((account_id, lp_coin.denom.as_str()));
+
+        // Iterate over all reward_denom keys
+        let keys_to_remove: Vec<String> =
```
piobab, line 265: Better to raise error for a key (not filter out)...

I would use:
`let keys_to_remove: StdResult<Vec<String>> = prefix.keys(store, None, None, Order::Ascending).collect();`
or
`let keys_to_remove = prefix.keys(store, None, None, Order::Ascending).collect::<StdResult<Vec<String>>>()?;`

contracts/incentives/src/astro_incentives.rs
```diff
@@ -232,29 +236,75 @@ fn decrement_staked_lp(
...
+
+    let new_total_staked_lp_amount = total_staked_lp_amount.checked_sub(lp_coin.amount)?;
+
+    // If the new amount is zero, remove the entry and all associated incentive states
+    if new_total_staked_lp_amount.is_zero() {
+        ASTRO_TOTAL_LP_DEPOSITS.remove(store, lp_key);
+
+        // Get all incentive states for the lp_key
+        let prefix = ASTRO_INCENTIVE_STATES.prefix(lp_key);
+        let keys_to_remove: Vec<_> =
```
piobab, line 297: Same here. See comment for line 265

piobab (changes requested):

piobab:
@markonmars remember about this one:
telegram-cloud-photo-size-4-5776364201894331878-y

markonmars:
closing, replaced with https://github.com/mars-protocol/contracts/pull/437

PR #439, opened 2024-12-17
MP-2702. Migration for Osmo contracts

piobab:
* Prepare migrations to align Osmo contracts with release https://github.com/mars-protocol/contracts/releases/tag/v2.1.0
* Fix missing connection in `account-nft` contract for rewards-collector contract


contracts/incentives/src/migrations/v2_1_0.rs
```diff
@@ -0,0 +1,43 @@
...
+        USER_UNCLAIMED_REWARDS_BACKUP.clear(deps.storage);
+    }
+}
+
+pub fn migrate(mut deps: DepsMut) -> Result<Response, ContractError> {
+    // Make sure we're migrating the correct contract and from the correct version
+    assert_contract_version(deps.storage, &format!("crates.io:{CONTRACT_NAME}"), FROM_VERSION)?;
+
+    // Clear old state
+    v1_state::clear_state(&mut deps);
```
markonmars, line 35: Is it possible there is still unclaimed incentives in this state?

markonmars (approved):

contracts/incentives/src/migrations/v2_1_0.rs
piobab: It is already migrated in previous migration. See our latest migration on Osmosis https://github.com/mars-protocol/contracts/releases/tag/v2.0.0

See here https://github.com/mars-protocol/contracts/blob/v2.0.0/contracts/incentives/src/migrations/v2_0_0.rs

piobab:
I pushed one commit with changelog update

bobthebuidlr (approved):

PR #441, opened 2025-01-10
Fix incentives config migration

piobab:


contracts/incentives/src/migrations/v2_1_0.rs
```diff
@@ -1,16 +1,28 @@
...
+    use cosmwasm_schema::cw_serde;
     use cosmwasm_std::{Addr, Decimal, DepsMut, Uint128};
-    use cw_storage_plus::Map;
+    use cw_storage_plus::{Item, Map};
+
+    #[cw_serde]
+    pub struct Config {
+        pub address_provider: Addr,
+        pub max_whitelisted_denoms: u8,
+        pub mars_denom: String,
```
piobab, line 22: `mars_denom` is in previous Config state and can't be deserialized to new Config type.

markonmars (approved):

contracts/incentives/src/migrations/v2_1_0.rs
Jloshow: I really need a life raft and some working room

Jloshow (changes requested):
Complete

PR #445, opened 2025-02-20
Add revenue share to Rewards Collector

markonmars:
This PR aligns this repository with the work done here - https://github.com/mars-protocol/core-contracts/pull/3

Probably best to review by commit

- the first two commits cherry pick relevant commits from core-contracts 
- the following commits are adjustments to suit osmosis deployment (without perps)


packages/types/src/address_provider.rs
```diff
@@ -37,6 +37,10 @@ pub enum MarsAddressType {
     Swapper,
     /// Astroport incentives contract
     AstroportIncentives,
+    /// Perps contract
+    Perps,
```
piobab, line 41: To be remove

contracts/rewards-collector/base/src/contract.rs
```diff
@@ -295,113 +297,305 @@ where
...
+        // The price of the asset to be swapped, denominated in the output asset denom
+        let asset_out_denominated_price =
+            slippage_adjusted_asset_in_price.checked_div(asset_out_price)?;
+        let min_receive_lower_limit = amount_in.checked_mul_floor(asset_out_denominated_price)?;
+
+        if min_receive_lower_limit > min_receive {
+            return Err(ContractError::SlippageLimitExceeded {
+                denom_in: asset_in_denom,
+                denom_out: asset_out_denom,
+                min_receive_minumum: min_receive_lower_limit,
```
piobab, line 479: typo `min_receive_minumum`

If we don't fix, there are 3 guys in the line waiting to fix it :D

packages/types/src/rewards_collector.rs
```diff
@@ -38,30 +66,26 @@ pub struct Config {
...
-    pub source_port: String,
-    pub acc_fee: Vec<Coin>,
-    pub timeout_fee: Vec<Coin>,
 }
 
 impl Config {
     pub fn validate(&self) -> Result<(), ValidationError> {
-        decimal_param_le_one(self.safety_tax_rate, "safety_tax_rate")?;
+        let total_tax_rate = self.safety_tax_rate + self.revenue_share_tax_rate;
+        decimal_param_le_one(total_tax_rate, "total_tax_rate")?;
```
piobab, line 88: I just notice that would be nice to have validation if sum of the tax rates for 3 bucket is 100%

markonmars: The third bucket is the fee collector, which is just the remainder after the other two - e.g `1 - (safety_fund + revenue_share)`, so it should always equal 1 for all three buckets regardless of how they are set here?

contracts/rewards-collector/base/src/contract.rs
markonmars: Will do this in another PR so it can be cherry picked to core-contracts - wdyt?

markonmars: https://github.com/mars-protocol/contracts/pull/446

packages/types/src/rewards_collector.rs
piobab: Yeah make sense. There are only two tax rates params:
```
    pub safety_tax_rate: Decimal,
    /// Percentage of fees that are sent to the revenue share
    pub revenue_share_tax_rate: Decimal,
```
So the third is the reminder.

bobthebuidlr (approved):

Jloshow (approved):

